PR #10, opened 2015-07-03
Scrolling performance

### LINKED ISSUE #3: Only show/render requests that are actually displayed
It would be better to show/render only requests that are actually visible for the user. This would be a massive performance improvement.

nerdbeere:
This PR fixes #3

### DIFF
--- a/src/component/main-content.js
+++ b/src/component/main-content.js
@@ -45,7 +45,7 @@ export default class MainContent extends React.Component {
 
   render() {
 
-    let {requests, config, services, activeWindow} = this.props;
+    let {requests, config, services, activeWindow, amountOfRequests, setFromIndex} = this.props;
     let {activeRequest} = this.state;
     let SearchBar, SetupInstructions;
 
@@ -88,6 +88,8 @@ export default class MainContent extends React.Component {
         requests={requests}
         config={config}
         setActiveRequest={setActiveRequest}
+        amountOfRequests={amountOfRequests}
+        setFromIndex={setFromIndex}
       ></Requests>
       {activeRequest}
     </div>
--- a/src/component/main-content.js
+++ b/src/component/main-content.js
@@ -47,8 +47,9 @@ export default class MainContent extends React.Component {
       });
     };
 
+    let activeRequestNode = null;
     if(activeRequest) {
-      activeRequest = <InspectRequest
+      activeRequestNode = <InspectRequest
         request={activeRequest}
         setActiveRequest={setActiveRequest}
       ></InspectRequest>;
@@ -77,7 +78,7 @@ export default class MainContent extends React.Component {
         setActiveRequest={setActiveRequest}
         setFromIndex={setFromIndex}
       ></Requests>
-      {activeRequest}
+      {activeRequestNode}
     </div>
   }
 }
--- a/src/component/main-content.js
+++ b/src/component/main-content.js
@@ -18,15 +18,6 @@ export default class MainContent extends React.Component {
     });
   }
 
-  _setFilter(str) {
-    if(str === '') {
-      str = null;
-    }
-    this.setState({
-      filter: str
-    });
-  }
-
   _setActiveRequest(request) {
     this.setState({
       activeRequest: request
@@ -45,14 +36,17 @@ export default class MainContent extends React.Component {
 
   render() {
 
-    let {requests, config, services, activeWindow, amountOfRequests, setFromIndex} = this.props;
+    let {
+      requestData,
+      config,
+      services,
+      activeWindow,
+      setFromIndex,
+      setFilter
+    } = this.props;
+
     let {activeRequest} = this.state;
     let SearchBar, SetupInstructions;
-
-    const setFilter = (str) => {
-      this._setFilter(str);
-      this.render();
-    };
 
     const setActiveRequest = (request) => {
       this._setActiveRequest(request);
@@ -66,7 +60,7 @@ export default class MainContent extends React.Component {
       ></InspectRequest>;
     }
 
-    if(requests.length > 0) {
+    if(requestData.totalCount > 0) {
       SearchBar = <Search setFilter={setFilter}></Search>;
     } else {
       SetupInstructions = <div className="setup-instruction">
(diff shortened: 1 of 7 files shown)


klipstein:
The container height is not adjusted when the list of requests got filtered. It should adapt to the amount of resulting items.

src/component/requests.js
```diff
@@ -15,32 +17,69 @@ export default class Requests extends React.Component {
     });
   }
 
+  componentDidMount() {
+    const requests = React.findDOMNode(this);
```
basecode: Wording: I see the variable name "requests" or some variation of it being used multiple times in that PR but often I just can't see the relation between the variable name and the data that the variable is holding.

I would rename to : scrollableDomNode or containerDomNode or listDomNode.

nerdbeere: Very valid point. To be honest I already thought about renaming this constant already

src/component/requests.js
```diff
@@ -15,32 +17,69 @@ export default class Requests extends React.Component {
...
+    requests.addEventListener('scroll', () => {
```
basecode: Do you remove that listener in `componentDidUnmount`?

src/component/requests.js
```diff
@@ -15,32 +17,69 @@ export default class Requests extends React.Component {
...
+      const fromIndex = Math.ceil(requests.scrollTop / requestElementHeight) - 15;
```
basecode: Have you checked how much time is spend on DOM manipulation in on frame while scrolling? I wonder if `scrollTop` triggers reflow. Think about using a Library that handles all that stuff. Listening on "scroll" usually is a performance killer.

src/component/requests.js
```diff
@@ -15,32 +17,69 @@ export default class Requests extends React.Component {
...
+      this._setFromIndex(fromIndex < 0 ? 0 : fromIndex);
```
basecode: Nit: Would rename to `_onScroll` or just get rid of `_setFromIndex` completely.

src/component/requests.js
nerdbeere: I think one performance issue might be calling react render in scroll. That's the only thing that seems to slow it down. I could debounce the scroll event. Do you have any suggestions?

src/component/requests.js
```diff
@@ -15,32 +17,69 @@ export default class Requests extends React.Component {
...
+    });
+  }
+
+  componentDidUpdate() {
+    const previousFilter = this.filter;
+    this.filter = this.props.requestData.filter;
+    const requests = React.findDOMNode(this);
```
basecode: Considering `componentDidUpdate` is called on every "scroll-frame", is `React.findDOMNode` a fast operation? Otherwise I would put it into the if condition (where you really need it).

nerdbeere: `componentDidUpdate ` is only called when the component was rerendered. So performance is not really an issue here.

basecode: Isn't the component re-rendered on scroll everytime as a result of `setFromIndex`?

src/component/requests.js
```diff
@@ -15,32 +17,77 @@ export default class Requests extends React.Component {
     });
   }
 
+  _onScroll() {
+    const scrollableDomNode = React.findDOMNode(this);
+    const fromIndex = Math.ceil(scrollableDomNode.scrollTop / requestElementHeight) - 15;
+    this._setFromIndex(fromIndex < 0 ? 0 : fromIndex);
```
basecode: `_setFromIndex` is a one-liner. Think about getting rid of it.

src/component/requests.js
nerdbeere: Yes but react won't update the component if nothing changed.

basecode: Actually I would call `React.findDomNode` only within the condition. I don't see it being used outside of the condition. Leads to less function calls, maybe even less DOM operations and increases readability.

src/component/requests.js
```diff
@@ -15,32 +17,77 @@ export default class Requests extends React.Component {
...
 
-    requests = requests.map((request) => {
+    if(requestData.filter) {
+      amountOfRequests = requestData.requests.length;
+    } else {
+      amountOfRequests = requestData.totalCount;
+    }
+
+
+    const requests = requestData.requests.map((request, index) => {
```
basecode: Have I mentioned that everything is called "requests" here? :)

src/component/main-content.js
```diff
@@ -45,15 +36,17 @@ export default class MainContent extends React.Component {
...
+
     let {activeRequest} = this.state;
     let SearchBar, SetupInstructions;
 
-    const filterRequests = (str) => {
-      this._filterRequests(str);
-      this.render();
-    };
-
     const setActiveRequest = (request) => {
```
basecode: There's already `_setActiveRequest`. Is that extra function necessary?

src/component/main-content.js
```diff
@@ -45,17 +30,21 @@ export default class MainContent extends React.Component {
...
-      this._setActiveRequest(request);
+      this.setState({
+        activeRequest: request
+      });
     };
 
     if(activeRequest) {
```
basecode: Please turn on http://eslint.org/docs/rules/no-shadow. Looks like "activeRequest" is used for a "request" object and also a "activeRequestComponent".

src/component/requests.js
nerdbeere: But this is actually an array of requests

src/index.js
```diff
@@ -81,14 +83,38 @@ const showWindow = (windowName) => {
   render();
 };
 
-function render(force) {
+const setFromIndex = (fromIndex) => {
```
basecode: Please add a comment. it's not clear to me what `setFromIndex` does.

src/component/main-content.js
nerdbeere: activeRequest gets wrapped in another object.

basecode: Could you check https://github.com/uxebu/james/pull/10/files#diff-ce2b02ac788b14260a1cae97ea32071eR51.  Either you're changing state within `render` or you shadowing a variable name.

src/component/requests.js
```diff
@@ -15,32 +17,73 @@ export default class Requests extends React.Component {
...
+
+    if(requestData.filter) {
+      amountOfRequests = requestData.requests.length;
+    } else {
+      amountOfRequests = requestData.totalCount;
+    }
+
 
-    requests = requests.map((request) => {
+    const requests = requestData.requests.map((request, index) => {
```
basecode: Please rename `requests` to `requestComponents` or sth along the lines. "requests" is just over-used.

PR #24, opened 2015-11-11
Add remove mapping to context menu

davidneat:


src/component/request.js
```diff
@@ -63,7 +63,7 @@ export default class Request extends React.Component {
   }
 
   render() {
-    const {request, response, showWindow, positionTop} = this.props;
+    const {request, response, showWindow, positionTop, urlMapper} = this.props;
```
nerdbeere: I would rather pass a function `removeMapping` instead of the whole urlMapper

src/component/request.js
```diff
@@ -84,6 +84,14 @@ export default class Request extends React.Component {
         showWindow('UrlMapping', {urlInput: request.fullUrl()});
         this._toggleContextMenu();
       }
+    }, {
+      title: 'Remove mapping',
+      icon: 'fa-trash-o',
+      onClick: (event) => {
+        event.preventDefault();
+        urlMapper.remove.call(urlMapper, request.fullUrl());
```
nerdbeere: I see no reason why you would need to use `call` here

PR #25, opened 2015-11-12
Add mapping active

davidneat:
https://github.com/uxebu/james/issues/19


davidneat:
tests for inactive mappings are currently missing

Issue #21, opened 2015-11-05
[Non-OSX] Items in top-bar menus simply error out


mitchhentges:
This includes any item under `James`, `Edit`, or `Window`
For example, upon clicking `James`>`About James`, the following window pops into the foreground:
![error](https://cloud.githubusercontent.com/assets/7784737/10977583/c7e8ba7e-83a6-11e5-84f6-0477c2c60ebf.png)

nerdbeere:
Thanks for your feedback.

mitchhentges:
This just in: this happens on Linux, too. I don't really know what I can describe about my environment that's relevant. Chrome version?

mitchhentges:
So, this issue seems to be because [`selector` in menu items is OSX-only](https://github.com/atom/electron/issues/2268). That's no fun :(

PR #27, opened 2015-12-23
Increase cross-compatibility

mitchhentges:
Replace "selector" with "role"
Make accelerator "CommandOrControl" for maximum compatibility

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,3 @@ build
 package
 node_modules
 binaries
-*.iml
-.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ build
 package
 node_modules
 binaries
+*.iml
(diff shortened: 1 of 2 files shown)


mitchhentges:
See [menu-item](https://github.com/atom/electron/blob/master/docs/api/menu-item.md)
and [accelator](https://github.com/atom/electron/blob/master/docs/api/accelerator.md)
for what these changes do

mitchhentges:
Oh, also fixes #21

.gitignore
```diff
@@ -2,3 +2,5 @@ build
 package
 node_modules
 binaries
+*.iml
```
nerdbeere: Could you please also remove this here?

mitchhentges: Derp, removed.

PR #36, opened 2016-01-21
Use browser-launcher2

mitchhentges:
* Now, launched browser will automatically work with proxy (didn't really before)
* Launches _new_ instance of browser
* Opens github page by default, rather than ""
* Make travis try packaging `james`, because sometimes that quietly breaks
* Removes a bunch of code. Ayy

### DIFF
--- a/src/util/open-browser.js
+++ b/src/util/open-browser.js
@@ -1,190 +1,20 @@
 /*eslint-disable */
-const chromeLauncher = require('karma-chrome-launcher');
-const firefoxLauncher = require('karma-firefox-launcher');
-const spawn = require('child_process').spawn;
-const tmp = require('tmp');
-const env = process.env;
-const path = require('path');
-
-import remote from 'remote';
-
-const app = remote.require('app');
-
+const browserLauncher = require('browser-launcher2');
 export default function openBrowser(browser = 'firefox') {
-
-  const proxy = 'localhost:1338';
-  const baseDecorator = function(self) {
-    self.id = browser;
-    self._getOptions = function(url) {
-      return [url]
-    };
+  const opts = {
+    browser: browser,
+    proxy: "localhost:1338"
   };
 
-  const launcher = {
-    chrome: {
-      create: function(params) {
-        var Executable = chromeLauncher['launcher:ChromeCanary'][1];
-        return new Executable(baseDecorator, params);
-      },
-      params: {
-        flags: [
-          '--user-data-dir=' + app.getPath('userData'),
-          '--proxy-server=' + proxy,
-          '--no-default-browser-check',
-          '--no-first-run',
-          '--disable-default-apps',
-          '--disable-translate'
-        ]
-      }
-    },
-    firefox: {
-      create: function(params) {
-        var Executable = firefoxLauncher['launcher:Firefox'][1];
-        return new Executable(browser, baseDecorator, params, {
-          create: function() {
-          }
-        });
-      },
-      params: {
-        prefs: {
-          'network.proxy.http': 'localhost',
-          'network.proxy.http_port': '1338',
-          'network.proxy.type': '1'
-        }
-      }
-    }
-  };
-
-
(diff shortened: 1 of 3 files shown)


src/util/open-browser.js
```diff
@@ -1,190 +1,20 @@
...
-    self._clearTempDirAndReportDone(error)
-  };
-
-  this._clearTempDirAndReportDone = function(error) {
-    tempDir.remove(self._tempDir, function() {
-      self._done(error)
-      if (onExitCallback) {
-        onExitCallback()
-        onExitCallback = null
+    launch('https://github.com/uxebu/james', opts, function(err) {
```
nerdbeere: I'd rather use a default page that is not https as long as https doesn't work out of the box

mitchhentges: Good call. Any recommendations? http://www.uxebu.com/?

mitchhentges:
Hmm, browserify is throwing a fit about including `browser-launcher2`. Looks like it's in the launcher's dependency `headless`, at:

```
//index.js:4
try {
  spawn = require('child-killer').spawn;
} catch(e) {
  spawn = require('child_process').spawn;
}
```

Browserify doesn't support this automatically according to https://www.npmjs.com/package/browserify-optional

mitchhentges:
Alright, fixed, you've solved this problem before by "excluding" the troublesome modules from `browserify` using `--exclude` or `-u`

src/util/open-browser.js
```diff
@@ -1,190 +1,20 @@
...
-    self._clearTempDirAndReportDone(error)
-  };
-
-  this._clearTempDirAndReportDone = function(error) {
-    tempDir.remove(self._tempDir, function() {
-      self._done(error)
-      if (onExitCallback) {
-        onExitCallback()
-        onExitCallback = null
+    launch('http://www.uxebu.com/', opts, function(err) {
```
nerdbeere: I'd open a local index.html we can fill with information about james.

mitchhentges: Unfortunately, then we'd need to run a local webserver. If not, then the file will just load, and the user won't see any action in James.
It's actually some accidentally UX, I think. Launching a browser both allows a user to get started, and also teaches them what James does by logging the launched browsers' requests. Oops, and hooray for seamless onboarding.

I think that it wouldn't be a good tradeoff to just load a file locally

mitchhentges:
... making Travis retry the build, because it's flaky for some reason

src/util/open-browser.js
nerdbeere: True, but I'm not sure if it's a good idea to open another port on the users machine.

mitchhentges: You don't want to open another port, like run a webserver locally, right? So that removes the option:
* ~~Open `localhost:8081` or whatever~~

What other options do we have? I have two that I can think of:
* Simply open a file, but not over http, e.g. `./james-info.html`. However, this won't show a network request in James, and we lose that onboarding
* Open a remote network resource, e.g. `http://???.com`. We'd want to open a page that has details about James, but unfortunately, the only one is the Github page, which is served over `https` and the user doesn't have immediate support for.

I'm personally leaning towards opening an arbitrary page over `http` to onboard (wrong term, but close enough) new users, showing them how James works. Once `https` is out-of-the-box, then we can default to the GitHub page?

Ooh, one other idea is to point towards `http://localhost:8000` and assume that users are doing local development on that (fairly standard) port

nerdbeere: We could open a page like `http://james.awesome` and use a default mapping to point this request to a local file. This would even better demonstrate the power of james I think

nerdbeere:
Can we merge this?

mitchhentges:
Sure, but it isn't doing that default URL mapping yet :)

nerdbeere:
That's fine. We should do this in a separate PR anyway.

mitchhentges:
#yolo let's do it ;)

PR #45, opened 2016-01-22
Slow connection

### LINKED ISSUE #17: Simulate slow connections (throttle)
As mentioned [here](https://twitter.com/wakkahari/status/620878858909294593) it would be a cool feature to implement a throttle feature to test a page with slow connections.
This should be fairly easy to implement using hoxys slow feature: http://greim.github.io/hoxy/#request-slow

Implementation could be done here: https://github.com/uxebu/james/blob/master/src/service/proxy.js#L50

mitchhentges:
Fixes #17.
Followed http://brewhouse.io/blog/2015/03/24/best-practices-for-component-state-in-reactjs.html quite closely, but it's still not great.

Feel free to be super aggressive in checking this code, I'm not familiar with React at _all_


src/service/proxy.js
```diff
@@ -110,4 +110,21 @@ export default class Proxy {
...
+   */
+  slow(rate) {
+    rate *= 1024;
+    this._proxy.slow({rate});
+  }
+
+  /**
+   * Disables throttling, allowing the proxy to run at maximum speed
+   */
+  relax() {
```
nerdbeere: can we find a better time for this?

mitchhentges: A better name? Probably, I'm up for suggestions, but the only others that came to mind were:
* open
* dethrottle
* clear
* flood

mitchhentges: Disablethrottling sounds good

src/index.js
```diff
@@ -154,7 +171,12 @@ function render() {
         isCachingEnabled={isCachingEnabled}
         requestData={requestData}
         clearRequests={clearRequests}
-        toggleCaching={toggleCaching} />
+        toggleCaching={toggleCaching}
+        onDisable={throttleDisable}
```
nerdbeere: this should be onDisableThrottle (onDisableThrottling?)

src/index.js
```diff
@@ -154,7 +171,12 @@ function render() {
...
+        onEnable={throttleEnable}
```
nerdbeere: this should be onEnableThrottle (onEnableThrottling?)

PR #49, opened 2016-01-23
Wildcards in URLs

### LINKED ISSUE #23: As a user I want to add mapping for a specific url structure
I want to add mapping rules with regex.

e.g.: cdn.com/*/index.js maps to local.js

cdn.com/version-1.0/index.js maps to local.js
cdn.com/version-2.0/index.js maps to local.js

mitchhentges:
Fixes #23


test/unit/url-mapper-spec.js
```diff
@@ -102,30 +102,111 @@ describe('url mapper', function() {
...
+      function set(mapping) {
+        urlMapper.set(
+          mapping.url,
+          mapping.newUrl,
+          isLocal,
+          isActive
+        );
+      }
+
+      set(plain);
```
nerdbeere: Could you set this directly in the test that?

mitchhentges: Oops, good call

src/url-mapper.js
```diff
@@ -4,18 +4,73 @@ export default class UrlMapper {
...
+
+        // TODO handle case where chunks aren't a wildcard, but contain one, e.g. "http://test.com/version*/app.js"
+        // Either properly handle it here, or validate it in UI
+        if (chunk !== urlChunks[i]) {
+          isMatch = false;
+        }
+      });
+      return isMatch;
+    }).sort(function(a, b) {
+      // Use the mapping with the least wildcards
```
davidneat: Should we introduce a function with proper naming, which may help us to get rid of the comment?

src/url-mapper.js
```diff
@@ -4,18 +4,73 @@ export default class UrlMapper {
...
+      const regex = /[^\*]/g;
+      const countDiff = a.url.replace(regex, '').length - b.url.replace(regex, '').length;
+      if (countDiff !== 0) {
+        return countDiff;
+      }
+
+      // Equivalent amount of wildcards, use the url that's the "least-wildcardy"
```
davidneat: Should we introduce a function with proper naming, which may help us to get rid of the comment?

src/url-mapper.js
```diff
@@ -4,18 +4,73 @@ export default class UrlMapper {
...
+        return false;
+      }
+
+      let isMatch = true;
+      chunks.forEach(function(chunk, i) {
+        if (chunk === '*') {
+          return;
+        }
+
+        // TODO handle case where chunks aren't a wildcard, but contain one, e.g. "http://test.com/version*/app.js"
```
davidneat: Maybe we should create an issue and remove the todo in the code.

src/url-mapper.js
```diff
@@ -4,18 +4,73 @@ export default class UrlMapper {
...
-    return this._map[url];
+    const plainUrl = this._map[url];
+    if (plainUrl) {
+      return plainUrl;
+    }
+
+    const wildcardList = [];
+    for (const key in this._wildcards) {
+      if (this._wildcards.hasOwnProperty(key)) {
+        wildcardList.push(this._wildcards[key]);
```
davidneat: Might the following also be an option?

const wildcardList = Object.keys(this._wildcards).map((key) => this._wildcards[key]);

mitchhentges: Way cleaner, thanks!

Issue #35, opened 2016-01-21
Show available browsers when James is launched


mitchhentges:
Right now, Chrome and Firefox are always shown, regardless of whether they are installed, or if other browsers are available.

Instead, `browser-launcher2`'s [detection feature](https://github.com/benderjs/browser-launcher2#browser-detection) should be used to find potential browsers

nerdbeere:
That's an idea we had some time ago: https://github.com/uxebu/james/issues/9
As you explained this a lot better, I'm gonna close #9

mitchhentges:
Oh, I thought #9 was strictly about using the other library. I was working on that when I had the idea about detection :wink:

PR #50, opened 2016-01-26
Show all available browsers when launching proxy

mitchhentges:
Just got all the logos set up:

screenshot from 2016-01-26 10-45-49


mitchhentges:
Fixes #35

Issue #37, opened 2016-01-21
Show an error when browser fails to launch


mitchhentges:
Currently, the user is provided no feedback when a browser doesn't launch. Instead, they get to wait to see if it eventually pops up.
Perhaps a modal should be shown? A toast, like in Android? A text log at the bottom? I'm not sure, but _some_ feedback should exist

PR #53, opened 2016-01-26
Show browser as disabled with error message if it won't launch

mitchhentges:
If you mouse over the disabled browser, there's a tooltip
screenshot from 2016-01-26 12-56-34


mitchhentges:
Some tooltip action
img_20160126_130141

mitchhentges:
Fixes #37

PR #57, opened 2016-01-27
Make packaging, all file copying occur through gulp

mitchhentges:
I've left `mocha` as-is, because moving it to grunt won't remove any dependencies.
Additionally, the `electron` version has been bumped, because I #yolo pretty hard.

Also, I'm not really sure if it's real or not, but I did a side-by-side comparison of clean `npm run package`, and I've found the following:

e4c09ad11447c144e45cf7cf8114f1f44e849c0f: ~46s
53e7299eab86556b38659d2171ffd6fcf0832891: ~30s

I only ran the tests once, so it could be a fluke. I wasn't expecting this, as there's no "already-compiled-file" optimizations going on. Whatever, I'm not complaining :wink:


mitchhentges:
This branch also ensures that using `electron-connect` does not occur on production builds using `gulp-useref`

Issue #64, opened 2016-01-28
Show default page when a browser is started by james


nerdbeere:
A default page should be shown in a browser that was opened by james.

PR #70, opened 2016-01-28
Remove links to old uxebu/james repo, replace with links to new repo

mitchhentges:


davidneat:
Maybe we should also discuss, which website launches, when openBrowser is called.  https://github.com/mitchhentges/james/blob/master/src/open-browser.js#L13

related issue: https://github.com/james-proxy/james/issues/64

nerdbeere:
I already started working on replacing the default page: https://github.com/james-proxy/james/pull/77

Issue #66, opened 2016-01-28
CMD + Q doesn't quit application on OSX


kokarn:
The standard CMD + Q command doesn't quit the application. 
In fact, I can't seem to find any way to actually quit the application without right-clicking the dock or killing the process.

davidneat:
fixed by #72

kokarn:
Thanks! You guys are really quick

PR #72, opened 2016-01-28
Replace close shortcut to use 'Q' instead of 'W'

mitchhentges:
It makes sense: it closes an application, not a window of the application

### DIFF
--- a/src/menu.js
+++ b/src/menu.js
@@ -61,7 +61,7 @@ function menuTempl() {
       },
       {
         label: 'Close',
-        accelerator: 'CommandOrControl+W',
+        accelerator: 'CommandOrControl+Q',
         role: 'close'
       },
       {


mitchhentges:
Fixes #66

src/menu.js
```diff
@@ -61,7 +61,7 @@ function menuTempl() {
       },
       {
         label: 'Close',
-        accelerator: 'CommandOrControl+W',
+        accelerator: 'CommandOrControl+Q',
```
davidneat: Maybe we should think also about our friends using a Windows. I found something like in the electron docs:

```
      {
        label: 'Close',
        accelerator: (function() {
          if (process.platform == 'darwin')
            return 'CommandOrControl+Q';
          else
            return 'Alt+F4';
        })(),
        role: 'close'
      }
```

nerdbeere:
I agree that this should work on all platforms

src/menu.js
mitchhentges: Just tested on Windows: `Alt+F4` already works.
In fact, the previous key mapping, `Ctrl-W`, also worked on all platforms. It was just `W` instead of `Q`, and was therefore less intuitive.

davidneat: great!

PR #77, opened 2016-01-28
[WIP] Open default site that is mapped to a local file

### LINKED ISSUE #64: Show default page when a browser is started by james

nerdbeere:
Things to do:
- [x] Create a nice looking default page `resource/example.html`
- [ ] Create the mapping just once the first time the application gets opened

fixes #64


src/open-browser.js
```diff
@@ -10,7 +10,7 @@ export default function openBrowser(browser = 'firefox', failCb) {
     if (err) {
       failCb(err);
     }
-    launch('http://www.uxebu.com/', opts, function(launchErr) {
+    launch('http://james.james', opts, function(launchErr) {
```
mitchhentges: `http://james.proxy`, maybe?

nerdbeere:
Still looking for a way how to find out if the application was started for the first time.
I guess I could just create a file in the application space but that seems to be a bad solution.

mitchhentges:
What about storing a Boolean property in nedb?

nerdbeere:
Yep, we could create a new collection called Configuration where we store values like this one.

.gitignore
```diff
@@ -3,4 +3,5 @@ package
 node_modules
 binaries
 root-ca.crt.pem
-root-ca.key.pem
\ No newline at end of file
+root-ca.key.pem
+resource/example-site
```
mitchhentges: Why not just immediately put the compiled site in `build` or `package`?

script/build-example-site.js
```diff
@@ -0,0 +1,22 @@
...
+}
+
+var template = fs.readFileSync(basePath + 'resource/example-site.template.html', {encoding: 'utf-8'});
+var readmeText = fs.readFileSync(basePath + 'README.md', {encoding: 'utf-8'});
+
+var exampleSiteHtml = template
+  .replace('<!-- REPLACE -->', marked(readmeText))
+  .replace('src="resource', 'src="https://raw.githubusercontent.com/james-proxy/james/master/resource');
+
+fs.writeFileSync(basePath + 'resource/example-site/index.html', exampleSiteHtml);
```
mitchhentges: Maybe this could be moved to gulp, so that the target (`basePath + target + '/example-site/index.html')` can be dynamic.
The `resources` folder is usually full of constants, I'm not sure that it makes sense to put temporary files in there

tomitm: Is gulp really the right tool for that? I'd suggest defaulting to npm scripts and using gulp when you really need to.

mitchhentges: It's still a little weird to have temp files in resources, no?

If possible, though, it would be awesome to solve it without using gulp

tomitm: In `/resources` specifically, probably. Maybe a `/tmp` folder or the like? I should go find out if there's some sort of standard practice for that...

mitchhentges: Thought about this further over the last week. By moving the temp files from `resource` to somewhere else, we might be adding unneeded complexity. What's wrong with this solution? An extra line in `.gitignore`?

It's probably OK to leave it in resources. :+1:

PR #83, opened 2016-01-30
Properly handle trailing slashes

### LINKED ISSUE #41: Slashes matter in URL Mappings
`http://jsframework.io` won't match a mapping for `http://jsframework.io/`
Trailing slashes should be ignored

mitchhentges:
Fixes #41

### DIFF
--- a/src/url-mapper.js
+++ b/src/url-mapper.js
@@ -1,5 +1,9 @@
 export default class UrlMapper {
 
+  static unslashUrl(url) {
+    return (url.endsWith('/') ? url.substring(0, url.length - 1) : url).trim()
+  }
+
   constructor(db, update) {
     this._db = db;
     this._update = update;
@@ -15,7 +19,7 @@ export default class UrlMapper {
   }
 
   get(url) {
-    url = (url.endsWith('/') ? url.substring(0, url.length - 1) : url).trim();
+    url = UrlMapper.unslashUrl(url);
 
     const plainUrl = this._map[url];
     if (plainUrl) {
@@ -65,7 +69,7 @@ export default class UrlMapper {
   set(url, newUrl, isLocal, isActive = true) {
     isLocal = !!isLocal;
 
-    url = (url.endsWith('/') ? url.substring(0, url.length - 1) : url).trim();
+    url = UrlMapper.unslashUrl(url);
     newUrl = newUrl.trim();
 
     const mappedUrl = {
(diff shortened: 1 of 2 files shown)


src/url-mapper.js
```diff
@@ -63,12 +65,7 @@ export default class UrlMapper {
   set(url, newUrl, isLocal, isActive = true) {
     isLocal = !!isLocal;
 
-    // fix the url, because hoxy will add a slash to urls without path..
-    if (!isLocal && newUrl.split('/').length === 3 && newUrl.indexOf('?') === -1) {
-      newUrl += '/';
-    }
-
-    url = url.trim();
+    url = (url.endsWith('/') ? url.substring(0, url.length - 1) : url).trim();
```
nerdbeere: Could you get rid of the code duplication here?
Maybe we can move that to a helper which we can properly test.

test/unit/url-mapper-spec.js
```diff
@@ -152,11 +106,43 @@ describe('url mapper', function() {
...
     });
 
+    it('matches, if no trailing slash', function() {
+      const noTrailing = {
+        url: 'http://foo.com',
+        newUrl: 'newUrl'
+      };
+
+      set(noTrailing);
+      check('http://foo.com', noTrailing);
```
nerdbeere: I don't really like those tests.
For me a unit tests should always end with one expect. These `set` and `check` functions make it much more complicating to read tbh.

test/unit/url-mapper-spec.js
```diff
@@ -152,11 +106,43 @@ describe('url mapper', function() {
...
+      check('http://foo.com', trailingSlashes);
+      check('http://foo.com/', trailingSlashes);
+    });
+
     it('matches wildcards', function() {
       set(oneWildcard);
       check('http://foo.com/1/baz', oneWildcard);
     });
 
+    it('doesn\'t match wildcard regardless of trailing slash or not', function() {
```
nerdbeere: `or` in the description is always an indicator that it should be two tests instead of one :)

mitchhentges: Ah, I think I understand. These are acting more as integration tests, checking both in-and-out, rather than just a single function on `url-mapper`?

mitchhentges:
Should be fixed!

nerdbeere:
Could you fix the linting issues?

mitchhentges:
Oops sure

Issue #42, opened 2016-01-21
Strip protocol and match URLs, regardless of https or http


mitchhentges:
Still allow users to copy-paste the entire URL into the "Enter the URL to map" field, but remove the protocol when the mapping is "added".
When requests arrive, attempt to match them, regardless of `http` or `https`

PR #84, opened 2016-01-30
Protocol shouldn't matter in URL mapping source

mitchhentges:


mitchhentges:
The destination still needs a protocol, so I'm a little on-the-fence for handling this screen:
screenshot from 2016-01-30 18-38-58

Maybe I'll just throw an `http(s)://` in front of every request mask (the left column)

kokarn:
I think that makes it clearer. Especially when you start to look at
subdomains and stuff like that

mitchhentges:
I'm pretty happy with it:
screenshot from 2016-01-31 16-02-12

kokarn:
Yeah, that looks good.

However, isn't there legitimate use-cases for mapping http and https to
different resources?
Not sure how the browser behaves with proxy redirects from http -> https
and vice versa.

mitchhentges:
> Not sure how the browser behaves with proxy redirects from http -> https and vice versa.

It's just a standard 301 redirect, IIRC.
If there _is_ a legitimate use-case for mapping `http` and `https` to different resources, then #42 should be an invalid issue. It doesn't seem like it, though

kokarn:
Hmm, yeah that should be the case. I'm not sure how a local host would handle insecure content and stuff like that depending on protocol when you do a redirect.. That's pretty much the only thing I can come up with that would matter tho. Maybe when you switch origins for say youtube embeds and stuff like that as well, but can't really remember anything that would actually break anything.

mitchhentges:
Going to wait until #83 is merged before I fix this ones tests

mitchhentges:
This has a _lot_ of test cleanup, with `urlMapper.set()` now defaulting both `isActive` and `isLocal` to `true`, allow unnecessary complexity in the tests to be removed

PR #86, opened 2016-01-31
Ensure consistent font-size on toolbar, regardless of width

### LINKED ISSUE #85: `close` tab now has a gap of whitespace (if window width < 992px)
So, #74 fixed the situation for when the window width is wider, but for some reason, something is changing around the `990px` mark. Now, instead of a problem existing on big windows, it exists on small windows.

The close tab should _always_ appear properly.

screenshot from 2016-01-31 16-18-06

mitchhentges:
Fixes #85

### DIFF
--- a/style/components/_inspect-request.scss
+++ b/style/components/_inspect-request.scss
@@ -10,10 +10,11 @@
 
   .toolbar {
     position: absolute;
-    top: -42px;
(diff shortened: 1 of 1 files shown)


mitchhentges:
This is resolved better by #89, closing

PR #88, opened 2016-01-31
Make `Esc` close current window

### LINKED ISSUE #73: "ESC" should close request-details panel
It's a habit to tap "ESC" to get out of sticky situations. Chrome allows the same thing: when in the "Elements" view with the lower-console-panel open, hitting "ESC" makes it go away

mitchhentges:
Fixes #73


nerdbeere:
I would rather use 'Esc' to close in-app windows like the mapping window or the request detail view.

mitchhentges:
That's exactly what it does, sorry if my terminology was bad

PR #89, opened 2016-02-01
Misc minor UI cleanup

tomitm:
@mitchhentges pointed me towards this project, so I had a look around!

Noticed a few things that could use fixing:

- At first, wasn't completely sure what the random :no_entry_sign: in the bottom right did. Hovered over... nothing. So, I added a tooltip! The rest of the footer didn't have tooltips either, and there were a bunch of toggles. Best not to assume everyone realises they're buttons. So, tooltip all the things!
image

- While I was in the footer, I realised it could take advantage of React 0.14's [stateless components](https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html#stateless-functional-components). I was going to change `Throttle` as well, since it seems like it should be stateless as well, but that wasn't immediately working. I'll try to poke at it later, since I believe all the state should be higher up.

- While inspecting a request, the toolbar that contains close has an invisible overlay that prevents you from clicking or scrolling on elements underneath. That's fixed as well.

- Speaking of the close button, I noticed it had some alignment issues (#85) and fixed that as well. Also gave it a bit of `margin` from the sides so it didn't look awkward.
image

- In the list of requests, you can't really tell which one you're inspecting. So when you're looking at a bunch of similar or even identical requests, you're sorta out of luck. No longer an issue, as the active request is now **bold**.
image

Hopefully this isn't too many fixes for a single PR, if so, I can split it up.

### DIFF
--- a/src/component/footer.js
+++ b/src/component/footer.js
@@ -1,11 +1,11 @@
 import React from 'react';
-import {CacheButton} from './footer/cache';
+import CacheButton from './footer/cache';
 import Throttle from './footer/throttle';
-import {RequestCount} from './footer/requestCount';
+import RequestCount from './footer/request-count';
 
 const {func, object} = React.PropTypes;
 
-export const Footer = (props) => {
+const Footer = (props) => {
   const {
     toggleCaching,
     isCachingEnabled,
@@ -27,3 +27,5 @@ Footer.propTypes = {
   clearRequests: func.isRequired,
   requestData: object.isRequired
 };
+
+export default Footer;
--- a/src/component/footer.js
+++ b/src/component/footer.js
@@ -1,43 +1,25 @@
 import React from 'react';
-import Throttle from './footer-throttle';
+import {CacheButton} from './footer/cache';
+import Throttle from './footer/throttle';
+import {RequestCount} from './footer/requestCount';
 
 const {func, object} = React.PropTypes;
 
-export default class Footer extends React.Component {
+export const Footer = (props) => {
+  const {
+    toggleCaching,
+    isCachingEnabled,
+    clearRequests,
+    requestData,
+    ...other
+  } = props;
 
-  render() {
-    const {
-      toggleCaching,
-      isCachingEnabled,
-      clearRequests,
-      requestData,
-      ...other
-      } = this.props;
-
-    let cachingButton = <span>
-      <i className="fa fa-circle-o"/>
-      Caching disabled
-    </span>;
-
-    if (isCachingEnabled()) {
-      cachingButton = <span>
-        <i className="fa fa-circle"/>
-        Caching enabled
-      </span>;
-    }
-
-    return <div className="footer">
-      <button onClick={toggleCaching}>{cachingButton}</button>
-      <Throttle {...other} />
-      <div className="request-count">
-        Requests: {requestData.filteredCount}/{requestData.totalCount}
-        <button onClick={clearRequests}>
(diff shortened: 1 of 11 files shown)


src/component/footer/requestCount.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+const {object, func} = React.PropTypes;
+
+export const RequestCount = ({requestData, clearRequests}) => {
+  const {filteredCount, totalCount} = requestData;
```
mitchhentges: Ooh, ES6 string-template thingers, nice

src/component/footer/requestCount.js
```diff
@@ -0,0 +1,18 @@
...
+  return <div className="request-count">
+    {`Requests: ${filteredCount} / ${totalCount}`}
```
mitchhentges: Ooh, ES6 string-template thingers, nice

mitchhentges:
:+1:, but a bigger PR, so I'll set @nerdbeere merge

src/component/footer/requestCount.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
```
mitchhentges: Could you name this file in `hyphen-case` rather than `camelCase`?

nerdbeere: +1

src/component/footer.js
```diff
@@ -1,43 +1,25 @@
 import React from 'react';
-import Throttle from './footer-throttle';
+import {CacheButton} from './footer/cache';
+import Throttle from './footer/throttle';
+import {RequestCount} from './footer/requestCount';
 
 const {func, object} = React.PropTypes;
 
-export default class Footer extends React.Component {
+export const Footer = (props) => {
```
nerdbeere: Doesn't it make sense to still export it as default here?

src/index.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import hoxy from 'hoxy';
 import TitleBar from './component/titlebar';
-import Footer from './component/footer';
+import {Footer} from './component/footer';
```
nerdbeere: If you export footer as default you wouldn't have to use destructuring here

nerdbeere:
Awesome, thanks for your contribution!

src/component/footer.js
mitchhentges: The syntax `export default const ...` isn't valid, according to https://esdiscuss.org/topic/why-is-export-default-var-a-1-invalid-syntax.

Instead, you'd need to:
```
const Footer = (props) => {
...
}

export default Footer

nerdbeere: Which would make more sense to me :)

mitchhentges: Ah, just making sure that you got that :+1:

src/component/footer/requestCount.js
tomitm: I should also point out the subtle `/footer/` folder addition - is that okay?

src/component/footer.js
tomitm: Yeah, I ran into the invalid part Mitch pointed out, but didn't realise there was a workaround... weird that `export default class` is valid yet `export default const` is not. Special case for class, I guess? ¯\\\_(ツ)\_/¯

Fixed, but now I want to go find out why that's the case...

mitchhentges:
Looks real :+1:. Give master a merge, though, so we can verify that the build won't fail.

Issue #90, opened 2016-02-01
Chrome Canary, Firefox Dev Edition, etc aren't shown


tomitm:
I have the canary versions of Chrome and Firefox installed. I noticed that `browser-launcher2` does detect them, but gives both browsers the same name.

That also results in React complaining that keys aren't unique:
> Warning: flattenChildren(...): Encountered two children with the same key, `.2:$`. Child keys must be unique; when two children share a key, only the first child will be used.

Not exactly sure which browser version I end up launching at the moment, as the list order seems to change a lot of the time as well.

tomitm:
Couple of options:
- Submit a fix to `browser-launcher2` to identify the names differently (*chrome-canary*), then update James to show the different icons.
- Add an additional logic to look at the path as well. For example, Chrome Canary has *Chrome SxS* in it's path. Seems a bit sketchy, though, as paths are likely not guaranteed.
- Display the browsers in a list instead, showing the name + version number. Even though paths aren't guaranteed, at least version numbers are reliable.

Lemme know which approach you want to take, and I'll implement it.

kokarn:
I'd say try to get browser-launcher2 to identify the names differently.
Should be in their best interest as well.

mitchhentges:
It would be awesome if `browser-launch` was more reliable at showing the browsers. At the same time, I think that putting the browsers in a nice UI list would be cool, too.
![aoz8kgx8pzknypz7z38n](https://cloud.githubusercontent.com/assets/7784737/12713668/483fe062-c8d0-11e5-98cf-d95e1f6d57d5.jpg)

tomitm:
This is fixed with #91.

Also, `browser-launcher` just launches, it's actually [`win-detect-browsers`](https://github.com/vweevers/win-detect-browsers) that does the actual detection. I filed an issue, but it doesn't sound like the author is too keen on the idea, preferring to keep things simple.

So, #91 has a pretty simple check for the browser name based on the command. In that case that the check doesn't manage to figure out the difference, the UI list update does at least show the version, so you can figure it out yourself.

Hope that works out for everyone, lemme know if you have any objections on the PR.

mitchhentges:
It's a little cheesy (SxS for detecting canary?) But it's a definite improvement.

PR #91, opened 2016-02-01
Display browsers as list

tomitm:
See #90.

image

TODO:
- [x] Display browsers as list
- [x] Look into identifying canary/dev separately
- [x] Fix issue with overflow if the window height is smaller than the list
- [x] browser-launcher uses browser name, instead of path?

### DIFF
--- a/src/component/main-content/browser.js
+++ b/src/component/main-content/browser.js
@@ -1,6 +1,6 @@
 import React from 'react';
 
-const {string, func} = React.PropTypes;
+const {object, func} = React.PropTypes;
 
 export default class Browser extends React.Component {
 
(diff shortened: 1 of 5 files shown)


src/component/main-content.js
```diff
@@ -53,14 +53,34 @@ export default class MainContent extends React.Component {
...
+          if (a.version > b.version) return 1;
+
+          return 0;
+        })
+        .map((browser) => {
+          if (browser.name === 'ie') {
+            browser.name = 'Internet Explorer';
+          } else if (browser.type === 'chrome' && /SxS/.test(browser.command)) {
+            browser.name = 'Chrome Canary';
+          } else if (browser.type === 'firefox' && /Developer Edition/i.test(browser.command)) {
```
tomitm: Using `browser.command` might not work if the user changed their installation path, but the design still provides a fallback to the version number, so the user can differentiate between the two installations even if their names are the same.

tomitm:
Two notes on this:
- Absolute positioning is giving me trouble with the layout right now (don't want to implement a height workaround with JS on this..)
- browser-launcher2 uses the browser name as a launch option? That's.. odd. Need to look into that.

tomitm:
Fixed the overflow issue!

Although I noticed that `browser-launcher` uses the *browser name* to launch the browser...

tomitm:
Looks like I can pass the name and version, so that's set! :tada: 

@mitchhentges or @nerdbeere can you have a look and merge?

mitchhentges:
:+1:

PR #92, opened 2016-02-01
Query params

### LINKED ISSUE #62: Show query parameters as a list in the request detail view
They query parameters should be displayed as a list in the request detail view.

mitchhentges:
Fixes #62 
screenshot from 2016-02-01 10-33-36


src/component/inspect/key-value.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+const {string} = React.PropTypes;
+
+export const KeyValue = ({_key, value}) => {
```
mitchhentges: Not super happy with the naming here, I'm open to suggestions

nerdbeere:
This can get really long now. I think we need to implement a scrollbar or something like this

mitchhentges:
Should that be part of this issue?

nerdbeere:
In my opinion it should have at least a very simple solution.
I want to avoid breaking the master

mitchhentges:
`overflow-y: scroll` set!

PR #93, opened 2016-02-01
Put "lint" task back in `package.json`

mitchhentges:
Sorry for jumping backwards here, but here's what I'm thinking:
* It makes sense to have compilations using `gulp` to take advantage of incremental compiling
* Having `package` in `gulp` is a good idea so that `gulp-useref` can filter dev-only code
* Cleaning works nicely in gulp, because it's platform-agnostic
* Watching is super-easy in gulp, why not use it

However, for things like `eslint` and `mocha`, they don't work incrementally. Putting them in `gulp` just adds unnecessary overhead. By leaving them in `package.json`, it makes things more simple

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -7,7 +7,6 @@ const changed = require('gulp-changed');
 const sass = require('gulp-sass');
 const sourcemaps = require('gulp-sourcemaps');
 const autoprefixer = require('gulp-autoprefixer');
-const eslint = require('gulp-eslint');
 const browserify = require('browserify');
 const source = require('vinyl-source-stream');
 const electron = require('electron-packager');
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -83,17 +82,6 @@ gulp.task('package', ['package-resources', 'package-browserify'], (done) => {
...
 });
 
-gulp.task('lint', () => {
-  return gulp.src(['src/**', 'test/**'])
-    .pipe(eslint())
-    .pipe(eslint.format())
-    .pipe(eslint.failOnError())
-});
-
-gulp.task('clean', () => {
```
mitchhentges: There's two `clean` tasks, this removes one of 'em

PR #95, opened 2016-02-01
Only run useref on html files

mitchhentges:


gulpfile.js
```diff
@@ -13,6 +13,7 @@ const source = require('vinyl-source-stream');
 const electron = require('electron-packager');
 const useref = require('gulp-useref');
 const electronConnect = require('electron-connect').server.create();
+Only const gulpif = require('gulp-if');
```
nerdbeere: "Only"? :)

mitchhentges: Mfw linux didn't alt-tab

PR #105, opened 2016-02-03
Breaking down components

tomitm:
*So, this is kind of a big PR, sorry! Instead of 4-5 PRs, figured this would be a lot smoother with just one. The diff view doesn't look too clear, so have a look at the [folder view](https://github.com/tomitm/james/tree/components).*

---

One thing I've noticed is that most components tried to do a lot during `render`. I've tried to break down the existing components into the smallest possible pieces. This should help with growth later and enable component reuse.

At the very bottom of React's [component docs](https://facebook.github.io/react/docs/reusable-components.html#stateless-functions) (oddly enough), Facebook recommends using stateless components as a default, and then presumably the "full" components when you need to deal with state and things like `shouldComponentUpdate`.

> In an ideal world, most of your components would be stateless functions because in the future we’ll also be able to make performance optimizations specific to these components by avoiding unnecessary checks and memory allocations. **This is the recommended pattern, when possible.**

Couple of notes:
- I've split components into folders, to make it easier to understand how the app is composed. I had to name the top-level files as `index.js` to be able to import the component by name, which results in being able to import `./component/main-content` instead of `./component/main-content/main-content.js`
- Fixed the context menu behaviour; you were able to open it multiple times. Fixed this by moving that state out of `Request` up into `MainContent` where there's already state for `activeRequest`.
  - Bonus that `Request` is state-free now, meaning 1 stateful `MainContent` instead of *n* stateful `Request`s, should make React pretty happy.
  - Still kept `Request` as a "full" component for `shouldComponentUpdate`. Wondering what the cost to render `Request` is now that there isn't much logic in it.

I made `Throttle` into a stateless component, but I'm a bit concerned, as that results in having to re-render everything, since there's a `data` object that contains application state outside of React's scope. React's docs do seem to suggest that rendering everything is okay (see [*A Note on Performance*](https://facebook.github.io/react/docs/multiple-components.html#a-note-on-performance)), but somehow that doesn't seem realistic to me.

Might want to use a base `<App />` with [`setState`](https://facebook.github.io/react/docs/component-api.html) to manage `data` so that React knows to perform a shallow merge update of components without needing to call `render()`. Although, if that's at the top level, it's probably the same effect as right now, I imagine, so this might be moot.

I'm also wondering if [Redux](http://redux.js.org/) would be worth exploring to help with this, or if you might find it overkill. Seems simpler than Facebook's own Flux, anyway. Still exploring options, as I'm still not 100% sure of the best React way to do this (my recent background has been in Angular).

- [x] Break down components
- [x] Organize into component folders
- [x] Figure out main component import (index.js)


kokarn:
Redux is really nice, especially for this kind of application. 
Have touched both redux and reflux and both are really nice but i'd suggest going with Redux as it seems to be the way forward.

nerdbeere:
@mitchhentges any thoughts about this?

mitchhentges:
I'll check this out in about six hours, I had a bunch of classes moved to
today

mitchhentges:
From the react docs:
>The good news is that JavaScript is fast
lmao.

I'm not a big fan of calling each component `index.js` inside it's own folder, that makes it harder to open each component in each IDE, and requires manual going-through-the-folder-tree. Why not have the component at its usage level, then its subcomponents in a folder?
```
src/
  index.js
  components/
    main-content.js
    footer.js
    inspect-request.js
    footer/
      throttle.js
      clear-requests.js
    inspect-request.js
      basketball.js
      super-jook-dunks.pdf.js

File: src/component/inspect-request/request-details.js [hunk 1]
```diff
@@ -0,0 +1,34 @@
...
+  const {request} = props;
+
+  return <div className="box-body">
+    <section>
+      Request URL:
+      <FullUrl request={request.request} />
+    </section>
+    <section>
+      Request Query Params:
+      <RequestHeaders headers={request.request.query} />
```
mitchhentges: Not an issue with this PR per se, but we shouldn't show "Request Query Params" if there are none. Should it be in it's own component, so it can return `<div></div>` if necessary?

tomitm: I was thinking of that, hadn't thought of the best approach yet. I was thinking of adding a label prop to `RequestHeaders`, so it can be handled in there.

```jsx
<RequestHeaders label="Request Query Params" headers={request.request.query} />
```

That's one thought, anyway. But definitely agree that we shouldn't show an empty section.

tomitm:
> I'm not a big fan of calling each component index.js inside it's own folder, that makes it harder to open each component in each IDE, and requires manual going-through-the-folder-tree. Why not have the component at its usage level, then its subcomponents in a folder?

I was originally hoping I could name them `footer/footer.js` instead of `footer/index.js` but that didn't work with the module loader, unfortunately.

I don't completely agree it's harder to open each component in the IDE though. `ctrl-p` still works, so it doesn't require manual going-through-the-folder-tree.

The benefit, which you lose out on with putting the index outside, is you have all your related sub-components together. You're going to end up looking for `<CacheButton />` really quickly once you realise that logic isn't in `<Footer />`. Or really, you just ended up looking in the footer folder in the first place, which makes this point moot.

Or you could just flatten everything like before. Not sure if you want to do that with ~20 components and growing, hence what prompted the folders. Or we ditch the eslint rule preventing multiple components to a file, so we can cut down on the file sprawl at least, since the components tend to be fairly small.

¯\\\_(ツ)\_/¯

tomitm:
Alright, importing as `component/footer/footer.js` is actually the best solution. Also should make sure we're future compatible with whatever the ES module loader spec ends up being (which I can't find a clear answer on atm).

As I was going through files making sure imports were correct, I fixed up a couple of `require` statements that should be `import` for consistency.

src/component/footer/throttle.js
```diff
@@ -2,54 +2,25 @@ import React from 'react';
...
+
+  const icon = enabled ? 'fa fa-circle' : 'fa fa-circle-o';
+
+  return <div className="throttle">
+    <button title="Toggle throttling" onClick={toggleThrottle}>
+      <i className={icon} />
+      Throttle to (kBps):
+    </button>
+    <input type="text" defaultValue={rate} onChange={onRateChange} />
+  </div>;
```
mitchhentges: Much prettier

src/component/inspect-request/request-headers.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+
+const {object} = React.PropTypes;
+
+const RequestHeaders = (props) => {
```
mitchhentges: Is it really a `RequestHeader`? Though? What about when it's query parameters? or response headers?
I mean, it's better than `KeyValue`, which is so generic it hurts.
Just saying, if a better name comes to mind, let's do it

src/component/main-content/main-content.js
```diff
@@ -15,10 +15,37 @@ export default class MainContent extends React.Component {
...
+
+  isActiveRequest(request) {
+    if (!this.state.activeRequest || !request) {
+      return false;
+    }
+    return this.state.activeRequest.request.id === request.request.id;
+  }
+
+  setContextMenuRequest(request) {
+    if (!request || request.request.id === this.state.contextMenuRequest) {
```
mitchhentges: If I'm setting the `contextMenuRequest` to the current `contextMenuRequest`, then clear the current the `contextMenuRequest`? This isn't super understandable

src/component/requests/full-url.js
```diff
@@ -0,0 +1,39 @@
...
+    maxUrlLength = 60;
+  }
+
+  let port = null;
+  if (request.port) {
+    port = <span className="port">:{request.port}</span>;
+  }
+
+  return <div className="complete-url">
+    <span className="protocol">{request.protocol}//</span>
```
mitchhentges: `//</span`?

src/component/requests/request.js
```diff
@@ -0,0 +1,113 @@
...
+const {func, object, number, bool} = React.PropTypes;
+
+export default class Request extends React.Component {
+
+  shouldComponentUpdate(nextProps) {
+    return this.props.request.id !== nextProps.request.id ||
+      this.props.positionTop !== nextProps.positionTop ||
+      this.props.isActive !== nextProps.isActive ||
+      this.props.isContextMenu !== nextProps.isContextMenu ||
+      this.done !== nextProps.request.done;
```
mitchhentges: I know that `shouldComponentUpdate` was there before, but is it _really_ needed? Why only on `Request`?

src/component/inspect-request/request-headers.js
tomitm: Yeah, `KeyValue` did hurt a bit. At the time I named this, this was before you added the query params feature, and then I discovered it upon rebasing. I kind of justified it to myself by saying you could *maybe* consider query parameters to be a request header. It's a bit flimsy logic, but it kinda works.

Maybe `RequestDetails`? I'm open to suggestions. Won't hurt to rename later either.

src/component/main-content/main-content.js
tomitm: Yeah, the toggle case. I'll go doc that better.

src/component/requests/full-url.js
tomitm: I... *uh...*

...apparently that was there originally. That's weird. I'm going to go fix that, good eye.

src/component/requests/request.js
tomitm: This bit originally got me when I was moving around the context menu, and then just as I was about to go insane, I remembered about it.

Technically, `Request` and all of it's sub-components are now state-free, and there's minimal logic in `render()`, so it doesn't seem like a bad idea. I presume it was originally there because `Request` had state and was a bit more complex than it is now, so it was beneficial to check beforehand, multiplied by 10+ instances depending on window size.

src/component/inspect-request/request-headers.js
mitchhentges: Let's do some `RequestDetails` action for this PR, it sounds pretty nice

tomitm: Oh wait, I've already used `RequestDetails` in the body of `InspectRequest`. That's not going to work. :confused:

mitchhentges: `RequestMetadata`? idk

mitchhentges:
I'm happy with this, I'd recommend running it to make sure that it works (of course), but I'll leave the final merge to @nerdbeere 'cause it's a big one.
:+1:

PR #108, opened 2016-02-03
Max height for inspect-request

### LINKED ISSUE #100: Max-height of inspect-request window should be percentage of overall height
If the James window is too small, it's impossible to close the inspect-request window. That window should be maximum 60% (?) of the app's height
oops

tomitm:
Sets max height of inspect-request to prevent it from taking over the whole window and not being able to close the view.

Fixes #100.

### DIFF
--- a/style/components/_inspect-request.scss
+++ b/style/components/_inspect-request.scss
@@ -64,7 +64,6 @@
   }
 
   .box-body {
-    max-height: 500px;
     overflow-y: scroll;
   }
 
--- a/style/components/_inspect-request.scss
+++ b/style/components/_inspect-request.scss
@@ -7,6 +7,7 @@
   bottom: 0;
   z-index: 2;
   background: rgba($lightgray-light, 0.9);
+  max-height: 60%;
 
   .toolbar {
     float: right;


style/components/_inspect-request.scss
```diff
@@ -7,6 +7,7 @@
   bottom: 0;
   z-index: 2;
   background: rgba($lightgray-light, 0.9);
+  max-height: 60%;
```
mitchhentges: How does this behave, with `.box-body` still having a `max-height` of `500px`?

tomitm: Overlooked this originally...

`.box-body` will never get larger than 500px, which may constrain `.inspect-request` from ever reaching 60% height, depending on the screen/content/etc.

That'll affect fully maximized windows from ever reaching 60% (probably closer to 50%, depending on content), but when your window isn't full height (say, 400px height), this is what allows `.inspect-request` to hide the close button. With both rules in effect, the 60% on `.inspect-request` would then become the constraint, and `.body-box` may only reach a height of ~250px.

So, I'm going to go ahead and remove the 500px rule, since our intention here is to allow `.inspect-request` to reach up to 60% of the window.

PR #114, opened 2016-02-05
Material

tomitm:
- Quit pointing to CDN for Materialize and make it part of the build. This will allow for using `extend`, without having to pollute markup, and allows for fully local (offline) use now.
- Added depth to the header and search

image

**WIP:**
- [x] Header
- [x] Search

### DIFF
--- a/style/components/main-content.scss
+++ b/style/components/main-content.scss
@@ -19,14 +19,14 @@
     .search {
       @extend .card;
       margin: 10px;
-      padding: 0 15px;
+      padding: 0;
 
       input[type="text"] {
         margin: 0;
         border: 0;
         box-shadow: none;
         height: 25px;
-        padding: 5px 0;
+        padding: 5px 15px;
       }
     }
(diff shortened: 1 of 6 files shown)


tomitm:
Was considering putting search in the header, with that expanding search pattern, but I imagine it's better off as a prominent feature.

mitchhentges:
Don't forget to add it to `package.json`

style/components/main-content.scss
```diff
@@ -15,8 +15,20 @@
     left: 0;
     bottom: 0;
     right: 0;
-    height: 40px;
-    padding: 5px;
+
+    .search {
+      @extend .card;
+      margin: 10px;
+      padding: 0 15px;
```
mitchhentges: Can we move the padding to `input`? If a user clicks on what appears to be the search bar, but to the left of the input, the input won't be selected.
It should be possible to make the whole bar selectable

PR #115, opened 2016-02-05
Create new URL Mapping

tomitm:
I decided to split it up into a 2 part process, in order to properly highlight what's going on, and the option between mapping to a URL or local file.

The previous design also didn't exactly scale very well with a small width, so that's why I avoided relying on buttons at the end.

image

**URL to URL**
image

**URL to File**
image

*WIP*
- [x] Add from requests
- [x] Input validation
- [x] Keyboard usability

/cc @nerdbeere @mitchhentges


mitchhentges:
Upon opening "URL Mappings":
```
Warning: Failed propType: Required prop `target` was not specified in `NewMappingTarget`. Check the render method of `NewMapping`.
Warning: Failed propType: NewMappingTarget: prop type `createUrl` is invalid; it must be a function, usually from React.PropTypes. Check the render method of `NewMapping`.

**Comment by mitchhentges** [comment]
There's some other components that also complain about proptypes, I'd recommend creating both a URL and a file mapping to see them all.

Also, can you make the cursor over "Choose file" a pointer? Maybe also make it look less like a textfield: maybe make the border go all the way around it, with some `border-radius`. Later, we can add drag-and-drop:
screenshot from 2016-02-11 19-35-06

**Comment by mitchhentges** [comment]
I can't "tab" over to the "choose file" input, and hit "space" to open the dialog. I have to use my mouse :(

**Comment by tomitm** [comment]
Fixed up the `propTypes`.. where's spell check when you need it?

As for the file input, yeah, good call. Definitely will make more sense for drag-drop later. I did take a crack at using the file input type initially, to get the native drag-drop, etc, but the value can't be set via JS, so React doesn't work too well with it.

**Comment by mitchhentges** [comment]
When entering a URL to map **to**, enter doesn't hit `CREATE` automatically. It would be pretty cool if it did. This is probably a thing for another PR, though

File: src/component/mapping/new-mapping-destination.js [hunk 1]
```diff
@@ -0,0 +1,61 @@
+import React from 'react';
+import remote from 'remote';
+import createChooseFile from '../../service/choose-file.js';
+
+const {bool, string, func} = React.PropTypes;
+
+const NewMappingTarget = (props) => {
```
mitchhentges: This feels like it should be two different components: `FileMappingTarget` and `UrlMappingTarget`. It feels like a lot of if/else all the way down for whether or not it's a `local` mapping.

src/component/mapping/new-mapping.js
```diff
@@ -0,0 +1,103 @@
...
+
+  reset() {
+    this.setState(initialState);
+  }
+
+  render() {
+    let form;
+    const {step, target, destination, isLocal} = this.state;
+
+    if (step === Step.target) {
```
mitchhentges: There's got to be a cleaner way of doing this. It's so weird that we've got a component that either returns one component or another

mitchhentges:
I'm not super happy with the two-faced components, but I'd be down to get some @nerdbeere insight.
Also, can you merge master? #pls.

src/component/mapping/new-mapping.js
tomitm: That's not all this does, it also keeps track of the state of the form, which is pretty key.

src/component/mapping/new-mapping-destination.js
tomitm: Hmm, yes and no. I do agree about the `local` bit, not completely happy about that. Otherwise, I wouldn't put this logic up into `NewMapping`, I'd keep this component, but it'd still be an if/else on whether to use `FileMappingDest` or `UrlMappingDest`. Make sure you look at my response to your comment about `NewMapping`, just to clarify things.

And ha, I keep mixing up target and destination in my head, yeah that's a typo.

src/component/mapping/new-mapping.js
```diff
@@ -0,0 +1,110 @@
+import React from 'react';
+
+import Constants from '../../constants.js';
```
nerdbeere: `Constants` should be lower case here.

nerdbeere:
I'm fine with the two-faced components. Somewhere you need to make the switch between those two.

PR #116, opened 2016-02-05
Don't include electron-localshortcut code in repo, set up `keyboard`

mitchhentges:


mitchhentges:
I think that I deleted this by accident. It's still needed, so I we don't have that `electron-localshortcut.js` file

mitchhentges:
Confirmed, I closed and deleted this... for no reason

PR #118, opened 2016-02-07
Fix footer overlap

tomitm:
Overlooked this issue in #108, was too distracted by figuring out the effects of the two heights to realise I completely broke scrolling and had some overlap in the process. Yay for absolute positioning, which doesn't care about it's surroundings. :anguished: 

Unfortunately, for this to work out, there needs to be a height in `box-body` (which is why it was there earlier, duh) and it can't be a percentage, because that becomes irrelevant with absolute positioning.

So, as a fix for now, I'm calculating the percentage on window `resize`. There's similar shenanigans going on in `<Requests />`, which sets the height of `requests-inner`... neither are really optimal.

image

### DIFF
--- a/src/component/inspect-request.js
+++ b/src/component/inspect-request.js
@@ -6,6 +6,11 @@ const {func, object} = React.PropTypes;
 
 export default class InspectRequest extends React.Component {
 
+  constructor() {
+    super();
+    this._onResize = this._onResize.bind(this);
+  }
+
   _getMaxHeight() {
     return window.innerHeight * 0.6 - 100;
   }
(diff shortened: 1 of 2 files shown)


src/component/inspect-request.js
```diff
@@ -6,13 +6,32 @@ const {func, object} = React.PropTypes;
...
+      maxHeight: this._getMaxHeight()
     });
   }
 
+  componentDidMount() {
+    window.addEventListener('resize', this._onResize.bind(this));
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('resize', this._onResize.bind(this));
```
mitchhentges: Does this work? When you run `bind()`, you create a new function, so the old listener will still be there

mitchhentges:
:+1:

PR #119, opened 2016-02-10
Properly handle broken url mappings

nerdbeere:
We should handle broken url mappings to we don't end up with a completely broken application.
~~Might also be worth investigating if it's possible to add empty mappings.~~
Also don't allow adding of mappings with either a empty url or new url

### DIFF
--- a/src/url-mapper.js
+++ b/src/url-mapper.js
@@ -141,7 +141,7 @@ export default class UrlMapper {
     this._removeMemoryCopy(mappedUrl.url);
 
     // ignore broken url mappings
-    if(!mappedUrl.url) {
+    if (!mappedUrl.url) {
       return;
     }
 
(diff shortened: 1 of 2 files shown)


src/url-mapper.js
```diff
@@ -140,6 +140,11 @@ export default class UrlMapper {
   _addMemoryCopy(mappedUrl) {
     this._removeMemoryCopy(mappedUrl.url);
 
+    // ignore broken url mappings
+    if(!mappedUrl.url) {
```
mitchhentges: This solves app-startup, when the mappings are read from the db and put into memory.
However, this won't stop new mappings from being added to the db. I mean, in memory, we'll _never_ have a broken mapping, but the db will fill up with broken mappings that are never used

nerdbeere: Sure, that's something we still need to fix.

PR #120, opened 2016-02-11
Fix for protocol

tomitm:
Apparently hoxy returns protocol as `http:` with the colon, but not the slashes? Sure..

Feel like it should just be `http` but I guess hoxy's author had their reasons.

nb: Github's syntax highlighting is misleading; that's not a comment. Escaping as `\/\/` will show as-is.


mitchhentges:
Maybe make an intermediate variable?

mitchhentges:
:+1: lgtm, just going to wait for Travis

PR #123, opened 2016-02-11
Throttle works again!

### LINKED ISSUE #121: Throttle control no longer works
Upon attempting to enable or change the value of the throttle, the following error appears in the console:
```
Uncaught TypeError: Cannot read property 'slow' of undefined
```
Fortunately, this issue isn't on `1.4.1`

mitchhentges:
Fixes #121

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -88,7 +88,8 @@ const toggleThrottle = () => {
   render();
 };
 
-const throttleRateChange = (kBps) => {
+const throttleRateChange = (event) => {
+  const kBps = parseInt(event.target.value);
   data.throttle.rate = kBps;
   proxy.slow(kBps);
   render();
--- a/src/index.js
+++ b/src/index.js
@@ -88,8 +88,7 @@ const toggleThrottle = () => {
   render();
 };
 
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -88,7 +88,8 @@ const toggleThrottle = () => {
   render();
 };
 
-const throttleRateChange = (kBps) => {
+const throttleRateChange = (event) => {
+  const kBps = parseInt(event.target.value, 10);
```
tomitm: `event.target.value` should be contained within the component it came from, `throttleRateChange` shouldn't have to worry about it. [Example](https://github.com/tomitm/james/blob/new-mapping/src/component/mapping/new-mapping-target.js#L8)

src/service/proxy.js
```diff
@@ -9,7 +9,7 @@ export default class Proxy {
     this._update = update;
     this._isCachingEnabled = isCachingEnabled;
 
-    const proxy = createHoxy();
+    const proxy = this._proxy = createHoxy();
```
tomitm: Why not just use `this._proxy` below? ¯\_(ツ)_/¯

PR #126, opened 2016-02-12
Fix #125 (electron X11 can't handle registering command)

### LINKED ISSUE #125: Keyboard shortcuts (can) cause lots of errors at startup on Linux
_Sometimes_, on launch, there's a bunch of errors that appear. On Linux, it can be enough to crash X. Awesome.
img_20160212_161038

mitchhentges:
Instead of registering `Command+Alt+I`, I register `CommandOrControl+Alt+I`.


tomitm:
:+1: Aren't you supposed to be using `CommandOrControl` always anyway? As I imagine this shortcut wouldn't have worked on Windows either, which lacks a cmd key.

PR #127, opened 2016-02-12
Show proxy status in bottom-left

mitchhentges:
* Green if proxy is working properly
* Yellow if partially working (e.g. no `HTTPS`, but plain `HTTP` works)
* Red if not working
screenshot from 2016-02-12 19-07-33


tomitm:
Slightly concerned about the `o` being used here, as the rest of them are toggles.

What about a green checkmark for working+https, white checkmark for httponly and a :warning: for not working? For some people, http-only may not be a concern, hence why I'm not sure I'd use a warning indicator for that.

src/component/footer/footer.js
```diff
@@ -2,22 +2,16 @@ import React from 'react';
 import CacheButton from './cache.js';
 import Throttle from './throttle.js';
 import RequestCount from './request-count.js';
+import HttpsStatus from './proxy-status';
```
tomitm: `ProxyStatus`?

src/index.js
```diff
@@ -138,8 +150,8 @@ keyboard.register('Esc', closeWindow);
 keyboard.register('CommandOrControl+U', () => showWindow('UrlMapping'));
 keyboard.register('F12', devTools.toggle.bind(devTools));
 keyboard.register('Ctrl+Shift+I', devTools.toggle.bind(devTools));
-keyboard.register('Command+Alt+I', devTools.toggle.bind(devTools));
-keyboard.register('Command+Alt+U', devTools.toggle.bind(devTools));
+keyboard.register('CommandOrControl+Alt+I', devTools.toggle.bind(devTools));
+keyboard.register('CommandOrControl+Alt+U', devTools.toggle.bind(devTools));
```
tomitm: I'm guessing another branch found it's way into this one?

style/components/_footer.scss
```diff
@@ -28,4 +26,29 @@
...
+    &.working i {
+      color: #55dd33;
+    }
+
+    &.partial i {
+      color: #bbaa44;
+    }
+
+    &.offline i {
+      color: #FF4444;
```
tomitm: Checkout the selection in `_colors.scss`, instead of defining new colours.

src/index.js
```diff
@@ -57,10 +60,19 @@ const createHoxy = () => {
...
   }
 
-  return hoxy.createServer(opts).listen(config.proxyPort);
+  const hoxyServer = hoxy.createServer(opts);
+  hoxyServer.on('error', (event) => {
+    data.proxyStatus = 'offline';
+    data.proxyMessage = event.code;
+    if (event.code === 'EADDRINUSE') {
+      // TODO make custom window
+    }
```
tomitm: Something tells me a `render()` might be necessary here, otherwise it might not show up right away.

src/component/footer/proxy-status.js
```diff
@@ -0,0 +1,27 @@
...
+
+const ProxyStatus = ({proxyStatus, proxyMessage, proxyWindow}) => {
+  const icon = proxyStatus !== 'offline' ? 'fa fa-circle' : 'fa fa-circle-o';
+  const label = 'Proxy: ' + (proxyMessage ? proxyMessage : proxyStatus);
+  const classes = ['proxy-status'];
+
+  if (proxyWindow) {
+    classes.push('with-info');
+  }
+  classes.push(proxyStatus);
```
tomitm: Why not just `const classes = ['proxy-status', proxyStatus]` above, while you're at it?

src/index.js
mitchhentges: Yep, otherwise I wouldn't be able to test this version (would crash)

src/index.js
mitchhentges: This runs synchronously. Since some other variables aren't set up yet, calling `render()` at this point causes `undefined` errors :/

src/component/footer/proxy-status.js
mitchhentges: Hah, good point

src/index.js
tomitm: Isn't it still possible for hoxy to emit an error event later though?

style/components/_footer.scss
```diff
@@ -28,4 +26,30 @@
...
+      color: white;
+    }
+
+    &.offline i {
+      color: $grapefruit-dark;
+    }
+  }
+}
+
+.footer > *:not(:first-child) {
```
tomitm: That can probably be optimized... wildcard aside, should at least remove the `:not(:first-child)` and use a second rule for `:last-child` to override the margin.

So that results in `.footer > *` and `.footer > *:last-child`

src/index.js
mitchhentges:
```
if (later) {
    render()
} else {
    // don't
}

Comment (tomitm) [inline src/index.js]: Could be as simple as `if (!!render) { render(); }`, instead of keeping track of `later`.

File: src/component/footer/footer.js [hunk 7]
```diff
@@ -2,22 +2,16 @@ import React from 'react';
 import CacheButton from './cache.js';
 import Throttle from './throttle.js';
 import RequestCount from './request-count.js';
+import ProxyStatus from './proxy-status';
```
tomitm: You should opt to use `.js` on your file imports, just in case things change down the road, wrt module spec.

style/components/_footer.scss
```diff
@@ -28,4 +26,30 @@
...
+    &.with-info {
+      cursor: pointer;
+    }
+
+    &.working i {
+      color: $grass-dark;
+    }
+
+    &.partial i {
+      color: white;
```
tomitm: `$white`

src/index.js
mitchhentges: I suddenly grok. In the `on` part, which is asynchronous. `render()` added, thanks :)

tomitm:
image

Wondering if the clear requests button should be in front, like the other icons? Kind of seems off to me, but I'm not sure. ¯\\\_(ツ)_/¯

tomitm:
And since you're messing with the footer, probably should set `cursor: default`.

Other than those two things, both minor, looks good! :+1:

tomitm:
When width is ~710px, the proxy status becomes hidden. Although, I think there's a couple of further improvements that can be made in the footer to handle smaller widths, so this isn't a blocker for this PR and I'll create a new issue.

PR #128, opened 2016-02-13
De-absolute-ify a bunch of elements

mitchhentges:


mitchhentges:
I'm not sure how exactly to implement #10 without absolute positioning. Did it _actually_ affect performance? Shouldn't the browser automatically be performant, since the elements are off-screen?

mitchhentges:
I've reproduced the efficiency benefits of #10, while also only rendering the _maximum_ number of requests (not 50). After this branch is merged, everything should _look_ the same, but should be resizable/changeable without everything burning. Yay!

tomitm:
image

I scrolled pretty fast, it broke. :cry:

Seems a *little* janky? That might be beside it has to quickly render new stuff that wasn't there before and I'm scrolling fast than it can keep up. Maybe add a 3-4 request buffer on both sides?

Oh, even better:

image

Note the cursor and the hover highlight. :confused:

tomitm:
Pretty close though! Really looking forward to this, as issues stemming from absolute positioning are blocking me on #91.

tomitm:
Talked with @mitchhentges earlier and he wasn't able to reproduce. I almost wasn't able to again until I had 100+ requests.

Looking into it some more and the fun part is the requests are actually rendered by React, just not displayed. You can click on them, and the inspect window will appear, but the requests will still be blank. The hover style is not displayed, but hovering with devtools will highlight the individual elements.

Messing with the `overflow` property on `.request-inner` (in devtools) triggers the element to show, so probably running into CSS shenanigans here.

src/component/requests/requests.js
```diff
@@ -4,33 +4,42 @@ import Request from './request.js';
...
+    this.fromIndex = this.fromIndex < 0 ? 0 : this.fromIndex;
+    this.props.setFromIndex(this.fromIndex);
   }
 
-  componentWillUnmount() {
-    const scrollableDomNode = ReactDOM.findDOMNode(this);
-    scrollableDomNode.removeEventListener('scroll', this._onScroll);
+  _onResize() {
+    const height = ReactDOM.findDOMNode(this).clientHeight;
+    this.requestCount = Math.ceil(height / requestHeight) + 1; // +1 if partial requests elements at top and bottom
```
tomitm: This should be 2, as 1 only accounts for top.

mitchhentges:
@nerdbeere this is ready for a performance test, and then ideally we should be able to merge it in. @tomitm can you check and potentially :+1:?

style/components/_requests.scss
```diff
@@ -1,36 +1,20 @@
...
+  overflow-x: hidden;
+  height: calc(100% - 55px); //55px is height of search bar (header)
 
   .request {
-    position: absolute;
-    left: 0;
-    right: 0;
+    position: relative; //absolutely-positioned context menu needs to be attached to request
     border-bottom: 2px solid $lightgray-light;
+    padding: 2px 0;
```
tomitm: Probably should specify a `height` to `.request`. Right now, it seems to be derived from the margin on `.labels`.

tomitm:
So, I loaded up about 300 requests. The aforementioned scrolling issues are no longer a problem *while there are no requests being added*. :tada:

However, turns out that `master` also has an issue with scrolling lagging while requests are being added. Should create a new issue to deal with that, as that isn't a regression from this PR. Was this an issue when the request limiting was originally added, @nerdbeere?

Everything else is looking good.

style/components/_requests.scss
mitchhentges: Since requests didn't have a height before, is rather not add then to this pr if I don't have to

PR #130, opened 2016-02-15
Narrow footer

tomitm:
Wide
image

Throttle disabled
image

Narrow
image

- [x] Hide proxy status message when width too narrow.
- [x] Hide throttle speed when throttle is disabled.
- [x] Cleanup heights and spacing on footer elements.
- [x] Figure out border on bottom (something to do with the throttle `input`)


tomitm:
This one's ready as well. :tada: @mitchhentges @nerdbeere

PR #131, opened 2016-02-15
Refactor proxy status error message

nerdbeere:


src/index.js
```diff
@@ -42,7 +42,7 @@ const data = {
   cachingEnabled: false,
   throttle: {enabled: false, rate: 0}, // rate is in kBps
   proxyStatus: 'working',
```
mitchhentges: We could make 'online' part of the `proxyStatus` "enum". Then "online" is always used.
`proxyMessage` is if something is wrong, and there should be more description other than "failed" or "partial"

src/component/footer/proxy-status.js
```diff
@@ -23,7 +39,7 @@ const ProxyStatus = ({proxyStatus, proxyMessage, proxyWindow}) => {
 };
 
 ProxyStatus.propTypes = {
-  proxyStatus: oneOf(['working', 'partial', 'offline']).isRequired,
+  proxyStatus: oneOf(Object.keys(constants).map(key => constants[key])).isRequired,
```
nerdbeere: I accept that this might be discussable. Should I write all of them down?

src/index.js
```diff
@@ -60,16 +61,14 @@ const createHoxy = () => {
...
+    data.proxyStatus = constants.PROXY_STATUS_NO_HTTPS;
   }
 
   const hoxyServer = hoxy.createServer(opts);
   hoxyServer.on('error', (event) => {
-    data.proxyStatus = 'offline';
-    data.proxyMessage = event.code;
+    data.proxyStatus = constants.PROXY_STATUS_ERROR_GENERIC;
     if (event.code === 'EADDRINUSE') {
-      // TODO make custom window
```
nerdbeere: Please don't add Todos to the code. If there is something we should not forget todo create an issue for it.

src/component/footer/proxy-status.js
mitchhentges: I think what you have there is better, so we don't need to write down the same status twice. :+1: :)

src/component/footer/proxy-status.js
```diff
@@ -1,15 +1,31 @@
...
+  [constants.PROXY_STATUS_WORKING]: 'fa-check',
+  [constants.PROXY_STATUS_NO_HTTPS]: 'fa-check',
+  [constants.PROXY_STATUS_ERROR_ADDRESS_IN_USE]: 'fa-exclamation-triangle',
+  [constants.PROXY_STATUS_ERROR_GENERIC]: 'fa-exclamation-triangle'
 };
 
-const ProxyStatus = ({proxyStatus, proxyMessage, proxyWindow}) => {
+const labelMap = {
+  [constants.PROXY_STATUS_WORKING]: 'Online',
+  [constants.PROXY_STATUS_NO_HTTPS]: 'HTTPs disabled',
```
mitchhentges: Why the lowercase 's'?

src/index.js
mitchhentges: Oop, sorry, sounds good

src/constants.js
```diff
@@ -0,0 +1,6 @@
+export default {
+  'PROXY_STATUS_WORKING': 'working',
```
mitchhentges: Maybe this might be worth using [Symbols](https://hacks.mozilla.org/2015/06/es6-in-depth-symbols/) for?

src/component/footer/proxy-status.js
nerdbeere: I don't know. You are the english native speaker :D

src/component/footer/proxy-status.js
nerdbeere: I'm not sure. Code should be explicit. This isn't explicit.

mitchhentges: Does `Object.values()` work?

src/component/footer/proxy-status.js
mitchhentges: Let's go full `HTTPS`, if that sounds good

src/constants.js
mitchhentges: Err, maybe not. Symbols allow non-colliding names, but I guess that it's not like the keys here were going to collide anyway. Nevermind haha

mitchhentges:
:+1:

PR #134, opened 2016-02-15
Context menu doesn't go away when clicking in whitespace

### LINKED ISSUE #129: When context menu is open, [left|right] clicking anywhere else should close it
Just like when you right-click on say, your desktop, the context menu will go away (or move) when you left-click or right-click elsewhere.

**To reproduce:**
1. Make a single request
2. Right click on the request
3. Left click in the whitespace before
The context menu is still there, but it should've gone away after `3.`

mitchhentges:
Fixes #129

### DIFF
--- a/src/component/main-content/main-content.js
+++ b/src/component/main-content/main-content.js
@@ -8,16 +8,13 @@ const {func, object, arrayOf, shape, array} = React.PropTypes;
 
 export default class MainContent extends React.Component {
 
-  constructor() {
-    super();
-  }
-
-  componentWillMount() {
-    this.setState({
+  constructor(props) {
+    super(props);
+    this.state = {
       filter: null,
       activeRequest: null,
       contextMenuRequest: null
-    });
+    };
   }
 
   setActiveRequest(request) {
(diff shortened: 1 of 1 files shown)


src/component/main-content/main-content.js
```diff
@@ -8,16 +9,13 @@ const {func, object, arrayOf, shape, array} = React.PropTypes;
 
 export default class MainContent extends React.Component {
 
-  constructor() {
-    super();
-  }
-
-  componentWillMount() {
-    this.setState({
+  constructor(props) {
```
mitchhentges: Move state setting to constructor, as laid out here: https://facebook.github.io/react/docs/reusable-components.html#es6-classes

src/component/main-content/main-content.js
```diff
@@ -78,8 +81,9 @@ export default class MainContent extends React.Component {
         request={activeRequest}
         setActiveRequest={this.setActiveRequest.bind(this)} />;
     }
+    const clickHandler = this.onClick.bind(this);
```
tomitm: You should probably just move this up into the constructor.

PR #135, opened 2016-02-15
Hardcoded proxyStatus propType change to proxy-status

mitchhentges:


tomitm:
> 	Apparently I was the only guy who though Object.keys() was OK here

It'd have been fine if the constants module would be limited to just proxy status constants, but I suspect it'll be used for other constants as well down the road, so that's going to mess with the proptypes here.

:+1:

PR #145, opened 2016-02-25
Fix heights when version number is missing

### LINKED ISSUE #141: Browser list icons not aligned properly
Maybe it's because `browser-launcher2` is only detecting one browser? At any rate, our CSS isn't properly handling the browser list for me.

tomitm:
Fixes #141.


mitchhentges:
:+1:

PR #147, opened 2016-02-26
Add scrolling to request mappings and prevent the browser selection f…

nerdbeere:
…rom being over the window

This fixes/optimizes the following things:
- Browser selection was in front of the window when no requests were made
- Mappings are now scrollable
- Window looks a little bit better now
- Use materialize button to close the window instead of a custom one

### DIFF
--- a/style/main.scss
+++ b/style/main.scss
@@ -76,37 +76,17 @@ $windowHeight: 400;
 
 .window {
   background: $white;
-  height: calc(100% - 40px);
 
   box-sizing: border-box;
   z-index: 1000;
+  position: absolute;
 
-  margin: 20px;
-  padding: 10px;
-  border: 6px solid $lightgray-dark;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
 
   box-shadow: 0 0 30px;
-
-  .close-button {
-    display: block;
-    height: 33px;
-    width: 33px;
-
-    position: relative;
-    top: -16px;
-    right: -16px;
-    float: right;
-
-    z-index: 1000;
-    background-color: $lightgray-dark;
-    color: $darkgray-dark;
-    padding-left: 11px;
-    padding-top: 5px;
-
-    i {
-      padding: 0;
-    }
-  }
 }
 
 strong {
--- a/src/component/mapping/url-mapping-window.js
+++ b/src/component/mapping/url-mapping-window.js
@@ -23,14 +23,19 @@ const UrlMappingWindow = (props) => {
   });
 
   return <div className="window url-mapping-window">
-    <a href="#!" className="close-button" onClick={closeWindow}>
-      <i className="fa fa-remove" />
-    </a>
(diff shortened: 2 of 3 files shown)


style/main.scss
```diff
@@ -76,37 +76,17 @@ $windowHeight: 400;
...
   z-index: 1000;
+  position: absolute;
 
-  margin: 20px;
-  padding: 10px;
-  border: 6px solid $lightgray-dark;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
```
mitchhentges: It's back :(
@tomitm, I'm not sure the best way to explain this. Perhaps it's OK to use `absolute` here? Going to need your input

mitchhentges:
There's an issue where `13px` of the URL Mapping list is missing at the bottom, and it's just white.

mitchhentges:
Looks really nice, and the close button is a good touch. However, I'm not sure about that `absolute`

style/main.scss
tomitm: Yeah, that'll be fine.

tomitm:
I was thinking about something like this, so this is nice to see!

image
*screenshot of this change just for context in this thread*

Have a look at the [Material guidelines for dialogs](https://www.google.com/design/spec/components/dialogs.html). This window is a full-screen dialog.

full-screen dialog

Most notably, this gives the window it's own header, close button in the top left. With this, don't really need to worry about the main toolbar still being clickable but not doing anything.

Anyway, that idea aside, this is really good as-is. :+1:

tomitm:
Also, @mitchhentges, not seeing the 13px on Windows.

image

Everything seems to line up nicely.

style/main.scss
nerdbeere: I really thought about doing it differently. But I think it's supposed to be like that? :/

nerdbeere:
Looks kinda weird with that scrollbar.

style/main.scss
mitchhentges: I remember that the reason that I started de-absoluting everything was because I was working on a `window`, but it didn't need to be full-size. I needed more flexibility.

However, that was a couple weeks ago, and I might've been mistaken. We can leave this as-is, and, worst-comes-to-worst, someone can change it _if_ they need to change it (when we add more windows)

nerdbeere:
Made it look a lot better now. But there's still an issue with the scrollbar on OSX.

tomitm:
Scrollbar looks way better now, nice!

image

There is one slight issue when the window is too narrow, those buttons don't behave nicely.

mitchhentges:
Maybe that's why the padding was there? #justSmallIssuesExploding

mitchhentges:
Other than the weird scrollbar for `nerdbeere`, this seems to be working well. At any rate, an offset scrollbar is far better than the current behaviour: browser lists showing up over top of mapping windows.

Merging

PR #152, opened 2016-02-29
Update eslint

mitchhentges:
* Fix existing code (consistent return just got _real_ mad)


CHANGELOG.md
```diff
@@ -11,9 +11,6 @@
 
 ### 1.4.3
 
-#### User features
```
mitchhentges: Derp, no point in this in an existing release

PR #155, opened 2016-03-03
Cache apt, node_modules, electron on Travis

mitchhentges:


nerdbeere:
Is it a good idea to cache `node_modules`? It makes the build faster but we won't see errors that arise when a dependency was upgraded and breaks something.

mitchhentges:
I guess that that would kind of be a good thing. The build would break less often (wouldn't fall apart when there's a bad dependency update), and we developers would probably run into the issue before the build/release system

nerdbeere:
We don't. How often to you delete your `node_modules` folder? In my experience you only see those errors because your build systems runs into that error.

mitchhentges:
But that build speed though ;)
Maybe we should give it a shot, and then see if we run into problems? Worst comes to worst, we can add `npm update` to our pre-run script

nerdbeere:
But usually build speed isn't an issue.
For me it's way more important to find breaking issues than to reduce build speed.

mitchhentges:
If I add `npm update`, can we keep the cache? That 50% build time is helpful, especially in PR turnaround

nerdbeere:
We should use [`npm --depth Infinity update`](https://docs.npmjs.com/cli/update), then this would be OK I guess.

mitchhentges:
Sweet!

PR #159, opened 2016-03-04
URL list should show browser's requested URLs

mitchhentges:
* Also makes mapping tag tooltips work with file mappings


src/service/proxy.js
```diff
@@ -41,7 +41,13 @@ export default class Proxy {
     request.done = false;
     request.id = uniqid();
     request.started = new Date().getTime();
-    request.originalUrl = fullUrl;
+    request.original = {
+      fullUrl: fullUrl,
```
nerdbeere: ES6 shorthand could be used here

PR #167, opened 2016-03-24
Don't report errors to Sentry if in development environment

mitchhentges:
Additionally, open the dev tools automatically if in development environment


src/dev-tools.js
```diff
@@ -1,8 +1,11 @@
 import remote from 'remote';
 
 export default class DevTools {
-  constructor() {
+  constructor(initial) {
```
tomitm: Thinking this could probably be named better. Or, maybe done outside (e.g., in index). Not 100% sure which was is best. Then again, don't really want to pile on more stuff into index.

`initallyOpen`?

tomitm:
Minor nitpick, otherwise :+1:

Issue #139, opened 2016-02-19
Inform users with an outdated version about the latest release


davidspinat:
It would be nice to inform users about the latest available release version. Users with a packaged version may see a notification about a newer build.

I would like to use the following api endpoint to check if the user has an outdated version: https://api.github.com/repos/james-proxy/james/releases Maybe it would be also nice to provide a direct download link to the latest package.

tomitm:
Better yet... auto-update? :grinning:

davidneat:
Unfortunately neither https://github.com/EvolveLabs/electron-updater nor http://electron.atom.io/docs/v0.34.0/api/auto-updater/ seem to fit our current needs.

The `electron-updater` would need a running sever which provides an api with the correct `Squirrel schema`. The `auto-updater` depends on `npm packages`, which we might also publish using a private server or using the usual npm server. Tough I like the idea to seperate James into small npm packages. What is your opinion?

tomitm:
Interesting, npm could work, but ideally if we can keep using just GitHub releases, less to maintain...

https://github.com/jenslind/electron-gh-releases

This one [looks](https://github.com/jenslind/electron-gh-releases/blob/master/docs/2.x/how-to.md) like it requires minimal changes, just a new `auto_updater.json` file for OSX updates.

Can probably use a `postrelease` npm script so we can just call `npm release` to make it easy to manage updates.

Hopefully that makes sense, haven't looked at it too closely yet.

mitchhentges:
With #343 merged, users are now notified about new versions of James (and have the ability to auto-update)

PR #169, opened 2016-04-04
Add auto updater

davidspinat:
https://github.com/james-proxy/james/issues/139

The following needs to be resolved:
- [ ] Add and update `auto_updater.json` within travis release
- [ ] Disable in development environment

I added the module `electron-gh-releases`, which runs in the main process and will check for new releases on github. This check will be performed on startup and will only inform the user, when an update is available. Following window will appear: 

bildschirmfoto 2016-04-01 um 16 20 19

If the user suspends the update, this window will be closed and pop up next startup. In case the user performs the update, the main window will be closed and the following loading screen will be shown:

bildschirmfoto 2016-04-01 um 16 40 52

I'm not sure how to properly test the whole updater functionality, as we would need to actually release a new version of James. Also the `auto_updater.json` is currently missing in the repository. Any suggestions how to integrate the `auto_updater.json` updates nicely during the release on travis? What are your opinions?

### DIFF
--- a/resource/updater.html
+++ b/resource/updater.html
@@ -1,0 +1,26 @@
+<!DOCTYPE html>
+<html>
+<head lang="en">
+    <meta charset="UTF-8">
+    <title>James Proxy</title>
+    <link href="./james.css" rel="stylesheet">
+</head>
+<body>
+    <div id="update">
+        <div class="spinner">
+            <div class="bounce-1"></div>
+            <div class="bounce-2"></div>
+            <div class="bounce-3"></div>
+        </div>
+        <div class="update-container">
+            <h2>New Version of James is available.</h2>
+            <h3>Would you like to download and install the update?</h3>
+            <div>
+                <button id="update-yes" class="btn waves-effect waves-light">Yes.</button>
+                <button id="update-no" class="btn waves-effect waves-light">No, remind me later.</button>
+            </div>
+        </div>
+    </div>
+    <script src="./updater.js"></script>
+</body>
+</html>
--- a/src/electron-app.js
+++ b/src/electron-app.js
@@ -1,36 +1,69 @@
-const electron = require('electron');
-const app = electron.app;  // Module to control application life.
-const ipc = electron.ipcMain;
-const BrowserWindow = require('browser-window');  // Module to create native browser window.
-const localShortcut = require('electron-localshortcut');
+import { app, ipcMain as ipc, BrowserWindow } from 'electron';
+import localShortcut from 'electron-localshortcut';
+import GithubReleases from 'electron-gh-releases';
 
-// Keep a global reference of the window object, if you don't, the window will
-// be closed automatically when the javascript object is GCed.
+const updater = new GithubReleases({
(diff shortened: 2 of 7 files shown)


mitchhentges:
Maybe we should make an `npm` script so that `npm run release $version` will change `package.json`, `auto_update.json` and make a git commit and tag

src/electron-app.js
```diff
@@ -1,36 +1,69 @@
...
-// Keep a global reference of the window object, if you don't, the window will
-// be closed automatically when the javascript object is GCed.
+import { app, ipcMain as ipc, BrowserWindow } from 'electron';
+import localShortcut from 'electron-localshortcut';
+import GithubReleases from 'electron-gh-releases';
+
+const updater = new GithubReleases({
+  repo: 'james-proxy/james',
+  currentVersion: app.getVersion()
+});
```
tomitm: Should put these into [`constants.js`](https://github.com/james-proxy/james/blob/master/src/constants.js), especially for version, as it's used for raven as well.

tomitm:
The [electron-gh-releases](https://github.com/jenslind/electron-gh-releases) repo has a tip suggesting to use [electron-release](https://github.com/jenslind/electron-release) for generating `auto_update.json`, which even seems to take care of creating a GH release.

tomitm:
@mitchhentges [`npm version`](https://docs.npmjs.com/cli/version) is built into npm, so you'd only need to do `npm version minor` to bump the version.

You can tie into it with `preversion`, `version` and `postversion` scripts, so that would be a good place to drop `electron-release` into.

resource/updater.html
```diff
@@ -0,0 +1,26 @@
...
+            <div class="bounce-1"></div>
+            <div class="bounce-2"></div>
+            <div class="bounce-3"></div>
+        </div>
+        <div class="update-container">
+            <h2>New Version of James is available.</h2>
+            <h3>Would you like to download and install the update?</h3>
+            <div>
+                <button id="update-yes" class="btn waves-effect waves-light">Yes.</button>
+                <button id="update-no" class="btn waves-effect waves-light">No, remind me later.</button>
```
tomitm: May want to look at the [Material spec](https://www.google.com/design/spec/components/dialogs.html#dialogs-specs) for dialogs. Should flip the buttons around, so that the affirmative action (*yes*) is on the right, and use `btn-flat`.

Wondering if you should go with `Remind me later` and `Update now`, to even out the lengths so that the no action isn't so much bigger than yes.

tomitm:
Hey @davidneat, are you still working on this? If not, I'd be happy to finish up!

Also, @mitchhentges and I were talking about this on the Matrix chatroom a while back, just in case you missed that.

davidneat:
@tomitm thank you for taking up again. Unfortunately I haven't continued. Feel free to finish up. 😊

tomitm:
Since I've changed quite a bit recently (yay merge conflicts), I'm going to finish this in a new PR. Thanks for the start, @davidspinat!

Planning on landing auto-update next. 🎈

PR #170, opened 2016-04-08
Update React, Electron, and Travis' Node version

mitchhentges:
* Dependencies updated
* Set `NODE_ENV` to `production` in `gulp` for some speed boosts in React
* Deployed `electron` version is now tied to `package.json`
* Internal `DEV` constant is inferred from `NODE_ENV`


gulpfile.js
```diff
@@ -59,31 +85,30 @@ gulp.task('package-resources', ['default'], () => {
...
-gulp.task('package-main', ['default'], (cb) => {
-  exec('browserify --node --igv=false -e build/electron-app.js -o ./package/electron-app.js --im', browserifyCb(cb));
+gulp.task('package-main', ['default', 'package-init'], () => {
+  browserify('build/electron-app.js', browserifyOpts)
+    .bundle()
+    .pipe(fs.createWriteStream('package/electron-app.js'));
 });
 
 gulp.task('package', ['package-resources', 'package-render', 'package-main'], (done) => {
+  const manifest = JSON.parse(fs.readFileSync('package.json', 'utf8'));
```
tomitm: `var manifest = require('./package.json')` works too.

gulpfile.js
```diff
@@ -59,31 +85,30 @@ gulp.task('package-resources', ['default'], () => {
...
+  const electronVersion = manifest.devDependencies['electron-prebuilt'];
```
tomitm: What happens if this ends up being changed to use a caret, like the rest of the dependencies?

tomitm: You can probably* use the following:

```js
const electronVersion = require('electron-package/package.json').version;
```

\* not entirely sure if it's a good idea or not atm, seems okay

gulpfile.js
```diff
@@ -59,31 +85,30 @@ gulp.task('package-resources', ['default'], () => {
...
+    .bundle()
+    .pipe(fs.createWriteStream('package/index.js'));
 });
 
-gulp.task('package-main', ['default'], (cb) => {
-  exec('browserify --node --igv=false -e build/electron-app.js -o ./package/electron-app.js --im', browserifyCb(cb));
+gulp.task('package-main', ['default', 'package-init'], () => {
+  browserify('build/electron-app.js', browserifyOpts)
+    .bundle()
+    .pipe(fs.createWriteStream('package/electron-app.js'));
```
tomitm: `gulp.dest('package')`, or is there some reason you're using `fs`?

https://github.com/gulpjs/gulp/blob/master/docs/recipes/browserify-transforms.md

gulpfile.js
```diff
@@ -49,7 +70,12 @@ gulp.task('resources', () => {
   ]);
 });
 
-gulp.task('package-resources', ['default'], () => {
+gulp.task('package-init', () => {
+  mkdirp.sync('package');
```
tomitm: Not necessary when using `gulp.dest`

> Folders that don't exist will be created.
> \- https://github.com/gulpjs/gulp/blob/master/docs/API.md#gulpdestpath-options

gulpfile.js
```diff
@@ -59,31 +85,30 @@ gulp.task('package-resources', ['default'], () => {
...
-    console.log(stdout);
-    console.error(stderr);
-    cb(err);
-  };
-};
-
-gulp.task('package-render', ['default'], (cb) => {
-  exec('browserify --node -e build/index.js -o ./package/index.js --im', browserifyCb(cb));
+gulp.task('package-render', ['default', 'package-init'], () => {
+  browserify('build/index.js', browserifyOpts)
```
tomitm: Just a heads up, this isn't exactly the same as before. There's a difference between render and main - render has `--igv=false` omitted (*see lines 71 and 75 of original*)

Interestingly, the [default](https://github.com/substack/node-browserify#usage) is not even a boolean, but `__filename,__dirname,process,Buffer,global`.

gulpfile.js
mitchhentges: Sure, sounds still readable, but shorter. Deal

gulpfile.js
mitchhentges: Yeah, because otherwise I need to use `vinyl-transform` or whatever.
At the same time, since I'm forced to use `mkdirp` because of the way `fs.createWriteStream()` works, I might as well use `vinyl`

gulpfile.js
mitchhentges: Yeah, see previous comment about wanting to avoid a dependency if necessary. But yeah, I'll swap out `mkdirp` for `vinyl-xtreme-shenanigans`

gulpfile.js
mitchhentges: Yeah, I don't know why that was there. Removing `--igv=false` produced the same `js` file.
So, no problem

gulpfile.js
mitchhentges: Also, wrt carets, `electron` shouldn't use carets. The deployed version (specified in gulp) _has_ to be an explicit version. It's a good idea to use the same version for development as deployment, so might as well keep it as a single version.

Besides, there's been breakage in "bugfixes" for electron before (the `object-registry` "object of undefined" thinger)

mitchhentges:
Updated

tomitm:
Funny that we were talking about caret versions, because it turns out `electron-packager@6.0.1` was having issues, and gulp wasn't setup to report that error properly. Both issues are now resolved. :ok_hand:

PR #172, opened 2016-04-22
Upgrade electron-packager, electron, etc

mitchhentges:
Use documented version of openDevTools
Don't need to set background-color to $white, fixed in electron


tomitm:
I was going to write a sassy comment about using carets and lack of skrinkwrap, but the fact that `node-sass` broke the build here is sass enough for one PR.

mitchhentges:
Hantera det.
uqoshwc

PR #176, opened 2016-05-04
Invalid version in prod builds

### LINKED ISSUE #173: Raven reports invalid version for PROD builds
Looks like Electron [checks `package.json`](https://github.com/electron/electron/blob/master/docs/api/app.md#appgetversion) for version information. When we deploy a release, there's no package.json. Perhaps `electron` embeds version information another way?

Worst comes to worst, we can embed `package.json`? I wouldn't like to, though

tomitm:
Specify `app-version` property in `electron-packager` so that `app.getVersion()` returns the correct version.
Cleanup: constants, require/imports in electron-app.js
Fixes #173.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -9,7 +9,7 @@ const sourcemaps = require('gulp-sourcemaps');
 const autoprefixer = require('gulp-autoprefixer');
 const electron = require('electron-packager');
 const useref = require('gulp-useref');
-const electronConnect = require('electron-connect').server.create();
+const electronConnect = require('electron-connect').server;
 const gulpif = require('gulp-if');
 const browserify = require('browserify');
 const source = require('vinyl-source-stream');
@@ -44,6 +44,7 @@ gulp.task('css', () => {
 gulp.task('resources', () => {
   return es.merge([
     gulp.src('node_modules/font-awesome/fonts/**').pipe(gulp.dest('build/fonts')),
+    gulp.src('package.json').pipe(gulp.dest('build')),
     gulp.src('resource/**')
       .pipe(changed('build'))
       .pipe(gulp.dest('build'))
@@ -120,8 +121,10 @@ gulp.task('watch', ['default'], () => {
 });
 
 gulp.task('livereload', ['default'], () => {
-  electronConnect.start();
-  const reload = () => electronConnect.reload();
+  var server = electronConnect.create({path: './build'});
+  server.start();
+  const reload = () => server.reload();
+
   gulp.watch('src/**', ['js', reload]);
   gulp.watch('style/**', ['css', reload]);
   gulp.watch('resource/**', ['resources', reload]);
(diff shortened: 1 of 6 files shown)


tomitm:
Verified working on Windows, haven't checked Linux and OSX builds yet.

mitchhentges:
:+1:

mitchhentges:
Jk, no :+1: here.
[`electron-prebuilt` doesn't set a version for Linux](https://github.com/electron-userland/electron-packager/blob/master/docs/api.md#app-version), so Linux still reports Electron's version, not James'.

We should still set `app-version`, that's a good idea. Perhaps we should `envify` a version, as well?

tomitm:
Should be ready to go now, thanks for checking that!

Shame about `app.getVersion()`. Wonder if electron-packager (or electron) would consider that a bug.. although, considering the following, might be misconfiguration on our part:

> [**app.getVersion()**](https://github.com/electron/electron/blob/969a30fc3bf12dd4b5bad5e193e43a340a45ac8b/docs/api/app.md#appgetversion)
> Returns the version of the loaded application. If no version is found in the application's `package.json file`, the version of the current bundle or executable is returned.

Checks `package.json` first, and looking at `binaries\James-linux-x64\resources\app\package.json`, ours is:

```json
{
  "main": "./electron-app.js"
}
```

Going to see if I can figure out why that's happening, then maybe I can revert back to `app.getVersion()`.

tomitm:
Ah, that didn't take long to find... [this `package.json`](https://github.com/james-proxy/james/blob/master/resource/package.json) is why.

This package.json gets packaged up, and of course, it doesn't have a `version` value, so electron falls back to the executable's `appVersion`.

tomitm:
Alright, removed the `resources/package.json` file and setup gulp to copy in `./package.json` so it gets packaged properly. That should be the proper fix!

mitchhentges:
This breaks `npm start`, because `main` no longer links to the `build` folder.
This can be solved, by changing the `start` script to: `cd build && electron . && cd ..`, but if `electron` fails to start, then it'll leave the user in the `build` folder.

mitchhentges:
Confirmed to work in Linux, Mac, and (VM) Windows. Nice.

tomitm:
Yeah, hold on, I'll have to figure out a better solution for `package.json`. Think I might need to generate it during gulp.

Going to dig around and see how much of `package.json` gets used by electron, as so far, it sounds like the `main` and `version` fields are important, but what else? Haven't found clear docs yet.

Also, this part isn't true, at least on Windows:

> but if electron fails to start, then it'll leave the user in the build folder.

mitchhentges:
No, that leaving in the build folder would only occur of you used my "cd
and electron and cd  .." fix

tomitm:
Should be good now. :sparkles:

mitchhentges:
Version is correctly shown for Windows, Mac, and Linux. `npm run build && npm start`, `npm run development` and `npm run package` work nicely for all platforms, too. Nice work

PR #178, opened 2016-05-05
Don't insert __dirname as undefined in browserify

### LINKED ISSUE #174: James on Windows doesn't detect browsers ... but only in PROD
When James is run on Windows, it doesn't show any browsers in the browser list. However, this only occurs after being `npm run package`'d. When run using `npm start`, everything is OK.

__To reproduce:__
1. Start Windows (7 or 10 confirmed)
2. `npm run package`
3. Start the windows build in`binaries`. Both the 32bit and 64bit builds are affected

mitchhentges:
Fix #174

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -72,7 +72,6 @@ function gulpBrowserify(entry) {
       global: undefined,
       'Buffer.isBuffer': undefined,
       Buffer: undefined,
-      __dirname: undefined,
       __filename: undefined
     },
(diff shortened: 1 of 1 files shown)


mitchhentges:
Though this now makes the `browserify` build work, it doesn't work with Electron, and doesn't solve the underlying issue: [win-browser-detect uses external scripts](https://github.com/vweevers/win-detect-browsers/issues/29)

PR #179, opened 2016-05-06
Minor UI cleanup

tomitm:
- Fix crazy titlebar shadow.
- Include Lato font, to remove network dependency on Google Fonts.


mitchhentges:
:+1: Looks way nicer, and moving the dep to compile-time is sweet

PR #182, opened 2016-05-07
Directly use nice woff2 lato font

### LINKED ISSUE #181: Request "method" text is not vertically aligned
This is due to the `lato` font change, of using the `lato-font` package rather than downloading it from Google Fonts at runtime. It looks like the `lato-font` font has extra padding underneat it, at least on linux.

```
@import url(http://fonts.googleapis.com/css?family=Lato:300,400,700);
```
2016-05-07-174128_219x103_scrot

```
@import '../node_modules/lato-font/scss/public-api';
@include lato-include-font('light'); // 300
@include lato-include-font('normal'); // 400
@include lato-include-font('bold'); // 600
```
2016-05-07-174146_201x120_scrot

mitchhentges:
Should fix #181


tomitm:
Looking good! :+1: Wanted to avoid committing the fonts to the repo originally, but it's not that big of a deal.

PR #183, opened 2016-05-07
Redux

tomitm:
Notable changes:
- Migration to main process
  - `Proxy`, `UrlMapper`, menu setup, and browser detection were moved into main process; too much was going on in the renderer, which actually slowed down request loads quite a bit
  - Goal is to keep most things out of the renderer and prefer to do work in main (especially if it requires `remote`), unless it's UI-related
  - Moved `src/electron-app.js` to `src/main/index.js`
    - There's actually no need to run main process code through babel/browserify (except for using `require` instead of `import`), since it's a node environment that supports ES6, but that's a change to be made later
- React Router replaces `windowFactory`
  - `TitleBar` and `Footer` are extracted out of `MainContent` into an `App` container, which RR controls the children of
  - `Home`/`Welcome` split out of `MainContent` so the user isn't contained to a single browser session, and can launch as many browsers as they want. URL mappings are also shown the same way instead of the original modal
  - `src/containers` contains top-level route containers, leaving the smaller components that make up the containers to the original `src/component`
  - `TitleBar` uses Material-ish tab styling and shows which tab is active. `Requests` also have an empty state now
    - This should make it really flexible to add more features and views, such as blocking rules, options, etc.
- Redux
  - `index.js` is no longer the kitchen sink! Use redux `connect` to allow a component to use the redux store - _dispatch_ actions for mutations, and make use of _selectors_ kept in `src/reducers` to read store data
    - The benefits of using actions and selectors are it makes it easy to understand what your app does by having all actions in one place, and by using selectors, you can make changes to the store shape without having to worry about the components. This leaves components to simply display the data that's given to them and delegate user actions to redux
  - `MainContent` is now `RequestsContainer` and no longer maintains which request is active or has a context menu (that's in Redux now)
  - Since `proxy`/`urlMapper` are in the main process, their states are synced with Redux over IPC
    - All changes should be dispatched through actions, and middleware will send them over IPC to main
    - If `proxy`/`urlMapper` have changes, they dispatch a `SYNC` action over IPC, which a reducer will apply to the state. This allows components to just read their states via redux's `connect` and not worry about how that data gets there
    - Note that to avoid excessive GC, request sync doesn't send full responses. When a request is set as active, the full response is grabbed at that time for only the single request. Was also going to send a minimal request as well, but I realised the list practically showed the whole request anyway, so that was moot
- Misc
  - Tests now run under electron's renderer
  - Added a _starting_ status for the proxy, so that the initial state isn't _working_ in case it's not
  - Raven tracks actions and sends them along with error reports as breadcrumbs

screenshot


mitchhentges:
Wooh, this is a big PR. I'll CR it "soonish"

package.json
```diff
@@ -8,14 +8,14 @@
...
-  "main": "electron-app.js",
+  "main": "main/index.js",
   "scripts": {
     "build": "gulp",
     "start": "electron ./build",
     "watch": "gulp watch",
     "development": "gulp livereload",
     "lint": "eslint ./src ./test",
-    "test": "mocha --reporter dot --recursive -r setup-referee-sinon/globals test/unit --compilers js:babel-core/register",
+    "test": "electron-mocha --renderer --reporter dot --recursive -r setup-referee-sinon/globals test/unit --compilers js:babel-core/register",
```
mitchhentges: neat

src/actions/app.js
```diff
@@ -0,0 +1,28 @@
...
+export function init(data) {
+  return (dispatch) => {
+    dispatch({
+      type: INIT,
+      ...data
+    });
+    dispatch(push('/'));
+
+    if (constants.DEV) {
+      dispatch(toggleDevTools());
```
mitchhentges: Why not `dispatch(toggleDevTools)`
and
```
export function toggleDevTools() {
    devTools.toggle();
}
```

src/component/footer/footer.js
```diff
@@ -1,25 +1,15 @@
...
-  </div>;
-};
-
-Footer.propTypes = {
-  toggleCaching: func.isRequired,
-  isCachingEnabled: func.isRequired,
-  clearRequests: func.isRequired,
-  requestData: object.isRequired
-};
+const Footer = () => <div className="footer">
```
mitchhentges: Don't you still need those dank props?

mitchhentges: > export default connect(mapStateToProps, mapDispatchToProps)(CacheButton);

Mmmm

.travis.yml
```diff
@@ -11,6 +11,8 @@ addons:
     packages:
     - icnsutils
     - graphicsmagick
+before_script:
+- export DISPLAY=:99.0; sh -e /etc/init.d/xvfb start || true
```
tomitm: Hmm, there was a reason for the `|| true`. Should've left a comment, my bad. I'll go poke around and see if I can find out again.

src/actions/app.js
tomitm: There's a few reasons:
- `dispatch(foo())` is a redux-ism, where `foo()` is an _action creator_ that returns a _flux standard action_ (basically, an object with the property `type`). So, the `toggleDevTools()` in your comment wouldn't make sense with `dispatch`, as dispatch expects an action to pass to it's reducers.
- `init` itself is an action creator with a thunk. Normally an action creator returns a single action. This one returns a function, so I can group a number of actions from one action.
- The `toggleDevTools` action is also already used in other places (keyboard shortcuts). Making an action for it, rather than what you suggested allows other parts of the app to hook into the action. I'm keeping extensibility in mind for down the road for plugins (hyperterm is an epic example of this).

src/component/footer/footer.js
tomitm: To explain this, redux's store emits updates on changes and `connect`ed components subscribe to those updates, updating as necessary.

`connect` creates a higher-order component (HOC) which has a pretty solid `shouldComponentUpdate` to prevent any extra changes, so it's a pretty decent fit in this case, I think.

src/store/index.js
```diff
@@ -0,0 +1,36 @@
...
+    urlMapperMiddleware,
+    proxyMiddleware,
+    routingMiddleware
+  ];
+
+  if (constants.DEV) {
+    middleware.push(loggerMiddleware);
+  }
+
+  const store = createStore(
```
mitchhentges: Just return `createStore()`

.travis.yml
tomitm: `|| true` was to skip OSX, since Xfvb doesn't happen on there. Buut that's definitely a bad idea in case it fails on Linux, so I've made it better. 👌

mitchhentges:
:+1:, quite a few changes, and I don't grok everything. However, it's working pretty well on Linux, and it's putting James in a good direction

PR #188, opened 2016-05-12
Fix electron deprecations

mitchhentges:
Allegedly, [if you were on `v0.37.8`, you'd be notified of deprecated API usage](https://github.com/electron/electron/releases/tag/v1.0.0).
Welp, we didn't have any errors, but updating to electron `v1.0.0` wasn't a safe move. This fixes the broken bits

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -1,4 +1,5 @@
-const {app} = require('electron').remote;
+import { remote } from 'electron';
+const { app } = remote;
 
 export default {
   'PROXY_STATUS_WORKING': 'working',
--- a/src/constants.js
+++ b/src/constants.js
@@ -1,5 +1,4 @@
-import remote from 'remote';
-const app = remote.require('app');
+const {app} = require('electron').remote;
 
 export default {
   'PROXY_STATUS_WORKING': 'working',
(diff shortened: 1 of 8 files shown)


src/constants.js
```diff
@@ -1,5 +1,4 @@
-import remote from 'remote';
-const app = remote.require('app');
+const {app} = require('electron').remote;
```
tomitm: Oh cool, we no longer need to do `remote.require`? Can you split this into the following:

```js
import { remote } from 'electron';
const { app } = remote;
```

Just want to standardize on `import` syntax over `require` if we no longer need it.

tomitm: Also, I think I'm inconsistent with spacing within `{...}`. Anyone have a preference? Pretty sure we can enforce it with eslint.

tomitm:
Ready to merge this in once you've fixed that one comment. :+1:

mitchhentges:
Fixed!

src/keyboard.js
```diff
@@ -1,13 +1,13 @@
-const ipc = require('electron').ipcRenderer;
+import { ipcRenderer } from 'electron';
```
tomitm: Don't need to change this, but just so you know, you could've used `import { ipcRenderer as ipc } from 'electron'` to keep the name!

Issue #122, opened 2016-02-11
Switch from `materialize` to `mdl`


mitchhentges:
[MDL](http://www.getmdl.io/index.html) doesn't require `jQuery` for neat-o "wave" effects, and it's a lot smaller.
Additionally, the `materialize` library has been giving some grief, with the media queries changing values and making some elements out-of-line. Ideally, a simpler material-design css framework would simplify this.

tomitm:
MDL is definitely a nicer and more complete MD library. :+1:

nerdbeere:
Makes sense to me. I never really liked `materialize`.

tomitm:
There's also [Material UI](http://www.material-ui.com/) which is a set of React components.

kokarn:
I've done some work with Material UI, it's really nice and well built. Also under active development.

tomitm:
I'm prototyping something using Material UI at the moment, it *is* really nice. Really tempted to recommend going that route with James. I'll see how much I still like it after I'm done this. :wink:

tomitm:
So, the JS for mdlite "upgrades" elements, which seems to interfere with React's event handling (click, etc) for elements. So when you click on something... nothing happens. Using mdlite without the JS is a bit quirky (input placeholders, for example). I'm not sure if I'm doing something wrong, but it's crapping out on my fairly simple prototype, and I get the feel that MDL isn't really meant to be used with React.

Material UI seems nice, except apparently performance for list-items is terrible, so I'd be weary about using that.

I'd either go with something like MUI, or there seems to be a bunch of smaller material react component libraries. Or, just DIY without a library would work too.

mitchhentges:
When in doubt, diy is possible :)
It's not like we need to be 100% compliant with the material spec anyways,
we just need James to look good

kokarn:
I'd probably go with material-ui and then
https://github.com/orgsync/react-list for the lists :)

PR #189, opened 2016-05-12
Save space on install

mitchhentges:
It's looking like, to use `electron-builder`, we're going to have to give up on `browserify` and throw our `node_modules` into the final build with everything. Since our non-devDependencies are still ~65MB, this PR will start cutting that down.

`no-optional` will allow us to avoid downloading features that we aren't using, e.g. [`hoxy`'s `cheerio`](https://github.com/greim/hoxy/pull/83).
Moving `materialize` into `style` will save us ~25MB, but will also maybe make us more keen on fixing #122

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ addons:
 script:
   - npm --depth Infinity update
   - npm run test-ci
+before_install:
+  - npm install --no-optional
 before_deploy:
   - npm run zip-binaries
 deploy:
--- a/package.json
+++ b/package.json
@@ -23,7 +23,6 @@
   "dependencies": {
     "hoxy": "^3.2.0",
     "james-browser-launcher": "^1.2.1",
-    "materialize-css": "^0.97.5",
     "nedb": "^1.7.1",
     "raven-js": "^3.0.2",
     "react": "^15.0.0",
--- a/style/main.scss
+++ b/style/main.scss
@@ -1,7 +1,7 @@
 $fa-font-path: "fonts" !default;
 
 @import '../node_modules/font-awesome/scss/font-awesome';
-@import '../node_modules/materialize-css/sass/materialize';
+@import 'materialize-css/materialize';
 
 @import 'colors';
 @import 'fonts';
--- a/style/materialize-css/components/_buttons.scss
+++ b/style/materialize-css/components/_buttons.scss
@@ -1,0 +1,197 @@
+// shared styles
+.btn,
+.btn-flat {
+  border: $button-border;
+  border-radius: $button-radius;
+  display: inline-block;
+  height: $button-height;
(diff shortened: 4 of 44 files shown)


tomitm:
> It's looking like, to use `electron-builder`, we're going to have to give up on `browserify` and throw our `node_modules` into the final build with everything.

This seems really crazy, so I'm going to hold off on this until I can do a bit more research and figure out what we can do.

tomitm:
...huh. So, the example app, [`onshape-desktop-shell`](https://github.com/develar/onshape-desktop-shell) seems to use TypeScript, so that's neat, but it still seems to package `node_modules` as-is?

Welp, time to resurrect `bower`...

Still kind of unclear to me why `node_modules` needs to be there, and why we can't just drop a browserified `index.js` in there like we have now. I mean, the point of having 2 `package.json` files is so they can be compiled against different targets, but I don't see why that should have anything to do with how the app is packaged. Is something flying over my head here?

mitchhentges:
> james-browser-launcher also seems to have some devDeps under deps.

I can see that `devDependencies` is showing up under its `package.json`, but there's nothing out-of-the-ordinary in `dependencies`:
2016-05-13-124257_895x383_scrot

> We only really dependencies there if we have any native modules, which I don't think we have... so maybe we shouldn't worry about that right now, and just keep doing what we're doing.

Hmm, good point. I'll need to look into `electron-builder` some more

> Last concern - seems like compiling for Windows from Linux isn't fully supported? Hence AppVeyor being used in tandem with Travis...

I'm making builds for windows successfully right now, using `electron-packager`. Besides, there's [docs for doing multi-platform on Linux for `electron-builder`](https://github.com/electron-userland/electron-builder/blob/master/docs/Multi%20Platform%20Build.md#linux)

tomitm:
Shouldn't `headless`, `mkdirp`, and `rimraf` be devDeps? Possibly others, haven't looked closely enough, but those stick out at me. Probably should discuss this in a jbl issue, but I think we can trim down our `node_modules` from there if we're going this route.

Those docs are A+, featuring stuff like "_Windows Code Signing doesn't work on Linux. We are going to fix it soon._" and "_Windows: Not documented yet._" :cry: Doesn't completely inspire confidence from me, but if it can work for us, why not.

But anyway, I think we might be able to keep our existing build setup, and then just use the `app` directory that `electron-builder` uses for compiled output.

mitchhentges:
Take a look at [the `electron-builder` branch in my repo](https://github.com/mitchhentges/james/tree/electron-builder), I'm most of the way there. Getting it to work with Travis might be a pain, though.
But yeah, this PR is probably not necessary

tomitm:
Yeah, that looks awesome so far! You should go ahead and open up a WIP PR for it. :+1:

PR #196, opened 2016-05-16
[WIP] Remove babel-preset-es2015

tomitm:
**WIP**

Remove `babel-preset-es2015` to minimize the transforms babel needs to make.
Unnecessary now that Electron ships Chrome 50 which supports most of ES6, and we have no other targets to compile for. :tada:

I have a feeling that I should bump up the version Travis' node runs on as well...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ matrix:
   - os: osx
 language: node_js
 node_js:
-- '5.1'
+- '6.2'
 addons:
   apt:
     packages:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,7 +8,7 @@ cache:
   - '%USERPROFILE%\.electron'
 
 install:
-  - ps: Install-Product node 5 x64
+  - ps: Install-Product node 6.2 x64
   - npm install
   - npm prune
   - npm --depth Infinity update
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -25,8 +25,12 @@ gulp.task('js', (done) => {
   return gulp.src('src/**')
     .pipe(newer('build'))
     .pipe(babel({
-      presets: ['es2015', 'react'],
-      plugins: ['transform-object-rest-spread']
+      presets: ['react'],
+      plugins: [
+        "transform-object-rest-spread",
+        "transform-es2015-function-name",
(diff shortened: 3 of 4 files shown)


tomitm:
https://kpdecker.github.io/six-speed/

On second thought, this might not the greatest idea yet. A few things are identical perf, but most things are a lot slower at the moment.

mitchhentges:
Personally, my two cents:
* Since we're using Electron, we're only targeting Node and Chrome - and modern versions, too
* It's not actually too much of a hit to use direct ES6 rather than Babel: there's quite a few things that are more performant with straight ES6
* Chrome/Node will continue to optimize straight ES6. At some point, we'll have a slower compile-time (due to Babel) _and_ a slower runtime (due to running ES6 as babel'd ES5). We can wait for Chrome and Node to improve further, but I'd rather remove compile-time complexity and let Electron upgrade automagically make things faster

tomitm:
> It's not actually too much of a hit to use direct ES6 rather than Babel: there's quite a few things that are more performant with straight ES6

I want to remove the babel plugins for ES6 features that are on par or exceed performance of ES5 equivalents. However, I want to keep the plugins for the ones where the ES6 equivalent is not yet as performant as the ES5 version.

> I'd rather remove compile-time complexity and let Electron upgrades automagically make things faster

Babel's always going to be part of our build thanks to React/JSX. The only complexity we'd be removing is a couple lines of babel config to include the plugins and their dependencies\*.

However, I really don't mind that our builds will take an extra 2-3 seconds on reload if that's going to impact our users with a **29x** hit to performance (see `spread` perf). I'm cherry-picking an extreme case, but I don't really like seeing 2x either. When we upgrade Electron, we should always watch for what kind of performance improvements we can take advantage of in new releases of Chrome, including removing babel plugins.

\* We can also simplify our `gulpfile` by removing the babel config out of there and keeping it in `package.json` (where it currently is) or even `.babelrc`. Not sure why it's duplicated across the two files right now.

tomitm:
Actually, I just realised something I didn't earlier. Babel plugins are only necessary for prod builds (to address the perf issues I said earlier) and we can remove them on dev (to address dev build concerns). Win-win! :tada:

tomitm:
Closing for now; will revisit this idea later.

PR #205, opened 2016-05-18
Don't recompile SCSS if not needed, remove unnecessary deps

mitchhentges:
* Use gulp-newer instead of gulp-changed, apply to css
* Don't autoprefix(), we're only targeting Node/Chrome

finger-guns


tomitm:
Don't have a chance to test this out yet, sorry, but why `gulp-newer` over `gulp-changed`? Seems similar, just different internal semantics.

Removing `autoprefixer` seems to make sense. Can't think of why we'd need it.

mitchhentges:
`gulp-newer` allows both "many-to-one" and "one-to-one" mappings, while `gulp-changed` is only "one-to-one". Since SCSS is many-to-one, JS is one-to-one, and I don't want _both_ `gulp-newer` and `gulp-changed`, I opted for just `gulp-newer`

tomitm:
Ah cool. I'll give it a test run tonight!

tomitm:
Can confirm, this is pretty slick. :+1:

PR #193, opened 2016-05-14
Electron builder

mitchhentges:
Same as #192, but a PR from the same repository so that I can test secure environment variables

PR #206, opened 2016-05-18
Use electric-squirrel for electron 1.0.0+ support

mitchhentges:
I forgot that for #193 I had manually changed `electron-builder` to work with `electron@^1.0.0`. This will use a compatible fork, because [mongo doesn't want to merge a PR](https://github.com/mongodb-js/electron-squirrel-startup/pull/3)

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "dist:win64": "build --platform win32 --arch x64"
   },
   "dependencies": {
-    "electron-squirrel-startup": "^0.1.4",
+    "electric-squirrel": "^1.0.0",
     "hoxy": "^3.2.0",
     "james-browser-launcher": "^1.2.1",
     "materialize-css": "^0.97.5",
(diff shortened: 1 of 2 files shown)


tomitm:
@mitchhentges Sounds like that PR got merged. Don't think this is still necessary, is it?

tomitm:
Looks like GK is on the ball with #209 and #210. :wink:

mitchhentges:
I suppose that we can stick with the original `electron-squirrel-startup`, though my fork has less dependencies, and less useless code. I'd prefer to go with the fork, due to less complexity, but I won't be too distraught to go with `electron-squirrel-startup` instead

tomitm:
It'd be best if you submitted a PR to get your changes in upstream, *ideally*.

Don't really see that much benefit to switching over to your fork, considering the big issue is fixed now and your changes are more of a nice to have sort of thing, which I don't think are worth having to maintain changes from upstream. :confused:

Just trying to save the maintenance hassle here. If upstream was abandoned, like browser-launcher was, sure. But this appears to still be maintained, and I think you're better off contributing your improvements upstream.

mitchhentges:
Sounds good

PR #220, opened 2016-06-04
Fix CSS watch

tomitm:
Most likely a #205 regression. Not really sure why this is no longer working, or maybe I didn't test this thoroughly enough originally.

Reproduce by changing the logo background-color in `_titlebar.scss` for complete obviousness.

Not completely convinced this is the right solution either, as it results in the following error during watch:

> Error in plugin 'sass'
Message:
    style\components\_browser.scss
Error: Undefined variable: "$mediumgray-light".
        on line 10 of style/components/_browser.scss
     border: 1px solid $mediumgray-light;
   ----------------------^

Which I know for sure is defined, in `_colors.scss`, so that leads me to think things aren't getting imported in the right order, or possibly everything's getting extra imported? Haven't looked closely enough there, as this solution seems wrong. Either that, or the current way of how we're doing sass is wrong (lack of imports except in main). Not 100% sure.

I figured an alternative solution would be:
```
  return gulp.src('style/main.scss')
    .pipe(newer({dest: 'build/james.css', extra: ['style/main.scss', 'style/**/*.scss']}))
```
but that's not working for me.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -34,7 +34,7 @@ gulp.task('js', (done) => {
 
 gulp.task('css', () => {
   return gulp.src('style/main.scss')
-    .pipe(newer('build/james.css'))
+    .pipe(newer({dest: 'build/james.css', extra: ['style/**/*.scss']}))
     .pipe(sourcemaps.init())
     .pipe(sass({outputStyle: 'compressed'}).on('error', sass.logError))
     .pipe(rename('james.css'))
@@ -42,7 +42,7 @@ gulp.task('css', () => {
     .pipe(gulp.dest('build'));
 });
 
-// There's two separate `resource-*` directories. 
+// There's two separate `resource-*` directories.
 // `resource-runtime` contains actual files which will exist in the same
 // directory as James at runtime, in production (e.g. the browser images).
 // `resource-compile` has resources that we don't necessarily want taking space with James, because they're compiled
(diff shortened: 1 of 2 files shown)


mitchhentges:
Looks like `newer` is only comparing the last-modified-date of `style/main.scss` to `build/james.css`, so if you change`_titlebar.scss`, `main.scss` is unchanged, and the build is skipped.

However, the entry point of `sass` _must_ be `main.scss`. If you pipe `**/*.scss` to `sass`, then you're going to have every file built separately, without imports (thus the error), and a build with all files (with `main.scss`).

This seems close:

```
 return gulp.src('style/main.scss')
    .pipe(newer({dest: 'build/james.css', extra: ['style/main.scss', 'style/**/*.scss']}))
```

mitchhentges:
The `extra` flag hasn't been released on `gulp-newer` yet, [I've opened an issue](https://github.com/tschaub/gulp-newer/issues/37) to hopefully get it published

tomitm:
> However, the entry point of `sass` _must_ be `main.scss`. If you pipe `**/*.scss` to `sass`, then you're going to have every file built separately, without imports (thus the error), and a build with all files (with `main.scss`).

Yeah, I figured that was the case. Didn't notice the `extra` flag wasn't released yet, nice catch! Looks like 1.2 is released now. :tada:

tomitm:
@mitchhentges Awesome, this looks like it works out great. Give it a spin!

PR #224, opened 2016-06-07
Get HTTPS certs from config location

mitchhentges:
This will require the [wiki](https://github.com/james-proxy/james/wiki/) being updated

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -21,6 +21,7 @@ import constants from './constants.js';
 
 import {remote} from 'electron';
 const {app, fs} = remote;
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -21,6 +21,7 @@ import constants from './constants.js';
 
 import {remote} from 'electron';
 const {app, fs} = remote;
+const userDataPath = app.getPath('userData');
```
tomitm: Probably should put this into constants, just like version?

jariangibson:
works like a charm!

PR #284, opened 2016-09-06
Automatic Updates

tomitm:
Ideally, automatic updates should be as minimally invasive as possible, so I've setup everything to run in the background and it should install on quit without too much user interaction.
- [x] Check for update on startup
- [x] Automatically download update if available
- [x] Show update status in footer
- [x] `auto_updater.json` for macOS (`npm version` management)
- [x] Link to release notes for platforms where auto-update is unsupported (linux)
- [x] Acquire a signing certificate; required for macOS releases
- [ ] Set `remoteReleases` to `true` (to generate update deltas)


tomitm:
Couple of initial notes:
- Unless we want to setup a [nuts](http://nuts.gitbook.com/) server or something, we're going to need to maintain an `auto_updater.json` file in master of [this format](https://github.com/Squirrel/Squirrel.Mac#update-json-format).
  - [`npm version`](https://docs.npmjs.com/cli/version) might be helpful to bump the version everywhere we need to (`package.json`, `auto_updater.json`).
- While Linux isn't support by Squirrel, `electron-gh-releases` at least allows Linux to check for updates. I'm not sure if we can provide auto-updates for Linux, but we can at least notify the user and link to the releases page.
- I noticed there's a `releaseNotes` property with the download, but it's not actually populated. Was thinking of showing something on the homepage when an update is ready to install, but maybe I'll just settle for [a link](https://github.com/james-proxy/james/releases/latest).
- Signed updates are required by Squirrel.Mac, but not Squirrel.Windows. Need to look into buying a certificate / Apple developer account / whatever.

tomitm:
This is pretty much done, save for setting up code signing. Have a look when you've got a chance, @mitchhentges 

I feel like I should tuck a note somewhere about creating new releases somewhere; should use `npm version <major/minor/patch>` from now on so that `package.json` and `auto_updater.json` are updated appropriately. Wiki might be a good place?

tomitm:
@nerdbeere Do you have an Apple Developer account? For auto-updates to succeed on macOS, we need a team "_Developer ID Application_" certificate.

Windows seems to not require signed updates, but Squirrel.Mac definitely requires it. Unfortunately I haven't found any way to work around this. :confused:

Documentation for code signing is [here](https://github.com/electron-userland/electron-builder/wiki/Code-Signing) - `electron-builder` takes care of everything, just need to export the certificate and set the `CSC_LINK` and `CSC_KEY_PASSWORD` env vars in Travis (probably best done within [repo settings](https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings)).

Also, for testing the macOS auto-updates, I can setup a quick test repo when we're ready, as `electron-gh-updater` is hardcoded to look for the `auto_updater.json` file on the specified repo's `master` branch.

Let me know if you want to chat about it, I can jump on Matrix/Riot/whatever with you! Bit difficult to develop/test for macOS when you're outside of the walled garden with Windows/Linux. 😅

nerdbeere:
yes, I have an totally unused apple developer account 😅
I will have a look tonight.

nerdbeere:
@tomitm I added both envVars to travis settings. Is there any way to see if this works?

screen shot 2016-09-20 at 19 38 33

tomitm:
@nerdbeere Awesome!! 🎉

So `electron-gh-releases` is [hardcoded](https://github.com/jenslind/electron-gh-releases/blob/master/src/GhReleases.js#L69) to search for macOS updates by checking an `auto_updater.json` file located at a [repo's](https://github.com/james-proxy/james/pull/284/files#diff-7b1ed8e005428717d211bea7f7e5c1feR20) `master`, which [contains](https://github.com/james-proxy/james/pull/284/files#diff-11cad47c787bb5331109d4abf816410bR4) the latest version and URL for the binary.

tl;dr - you'll need to do the following:
1. Create a new repo for testing
2. Add an `auto_updater.json` file [like this](https://github.com/james-proxy/james/pull/284/files#diff-11cad47c787bb5331109d4abf816410bR1), but change the version number in `name` and point `url` to a new **zip** that you've signed (auto-updates use the zip, not dmg)
3. In a test build, change [`repo` in `main/index.js`](https://github.com/james-proxy/james/pull/284/files#diff-7b1ed8e005428717d211bea7f7e5c1feR20), so that it points to this newly created repo

(alternatively, if you don't like the `auto_updater.json` shenangians, someone will need to host a [nuts](https://github.com/GitbookIO/nuts) server)

nerdbeere:
Any idea how I sign a `.zip` file? Am I using `electron-builder` for this?

tomitm:
I _think_ `electron-builder` takes care of it. Seems like it should be a matter of setting the environment variables then running `npm run dist`.

tomitm:
@nerdbeere Any luck? Lemme know if you want to chat on the Matrix channel or something.

nerdbeere:
Sorry for the delay, @tomitm 

I think the signing worked:
screen shot 2016-09-25 at 10 17 44

What I did to test this:
- created a new repository containing the `auto_updater.json`: https://github.com/james-proxy/auto-updater-test
- changed the Squirrel `enabled` flag to true in my local build
- created a new tag for the repository "auto-updater-test" that is named `v1.5.2`

Error I get:

```
[updater] Error when checking for updates: Error: Could not find a valid release tag.
    at /Users/hol0008j/Projects/james/node_modules/electron-gh-releases/GhReleases.js:156:17
    at process._tickCallback (internal/process/next_tick.js:103:7)
```

I looked into there and it gets `releases` as versions which is kind of odd.

nerdbeere:
Ok, I fixed this issue. The key was to actually *draft* a new release on Github: https://github.com/james-proxy/auto-updater-test/releases/tag/v1.5.4

Now I get the following error:

```
[updater] Error when checking for updates: { Error: Could not get code signature for running application
    at Error (native)
    at /Users/hol0008j/Projects/james/node_modules/electron-gh-releases/GhReleases.js:171:28
    at process._tickCallback (internal/process/next_tick.js:103:7)
  preventDefault: [Function: preventDefault],
  sender:
   AutoUpdater {
     _events: { 'update-downloaded': [Function] },
     _eventsCount: 1 } }
```

Seems like the current running application that wants to update also needs to be signed.

nerdbeere:
Now I opened the signed application and it seems to work:
screen shot 2016-09-25 at 11 06 41

but after restarting the app is just showing a blank screen :/
screen shot 2016-09-25 at 11 07 12

nerdbeere:
It worked! I uploaded a broken build (missing dependencies) to the release.

auto-updating-james

There is one issue left now:
After the restart it says "Unable to download update"

tomitm:
@nerdbeere 🎉🎉🎉 

I think the last problem might be because the name you're using in [`auto_updater.json`](https://github.com/james-proxy/auto-updater-test/blob/master/auto_updater.json#L2) doesn't match the version properly*, due to [this check](https://github.com/jenslind/electron-gh-releases/blob/master/src/GhReleases.js#L92-L96).

There should also be an error message when you hover over the status, which should tell you which error message it is. Hopefully if I'm right, it should be "_The feedUrl does not link to latest tag_..."

If that's the case, you should be able to just change `name` to `1.5.4` and once you relaunch, the error should go away and you shouldn't be shown an update is available.

\* In case you're wondering, to avoid this problem, I've automated this through use of `npm version foo` to automatically generate that file, so we don't have to worry about it going forward. Should take care of everything (except the changelog, which can be amended before push) for release.

nerdbeere:
screen shot 2016-09-26 at 07 41 10

This is the error I get. Will have to investigate further.
We could also release this and test it in production.

tomitm:
_Oh._ I was hoping it wasn't something like that... 😦

Digging around for that error in electron, found [this](https://github.com/electron/electron/blob/01f92a83ab83001a48644a32e2cfcfe569ef13c6/spec/api-auto-updater-spec.js#L24) - test for "emits an error on macOS when the application is unsigned". The error itself actually comes from Squirrel.Mac [here](https://github.com/Squirrel/Squirrel.Mac/blob/e9e2188cda3efb4bc08b1719bdef71880f9dc9b1/Squirrel/SQRLUpdater.m#L159).

So it sounds like you originally had a signed app, which updated successfully and then updated to an unsigned app? Uhh. :confused:

Maybe I missed something for making sure the update zips are signed?

nerdbeere:
I looked into it a bit more but can't get to the bottom of this.
The zips are definitely signed.

tomitm:
@nerdbeere 😦

One last final guess, try reverting [this change](https://github.com/james-proxy/james/pull/284/files#diff-94f6f38a9de3a246cc267c18b484384bR5). On Windows, James shows up under a folder with your name rather than _James Proxy_ because of that, which I think is confusing and not very intuitive.

If that doesn't work out, I think I'll just scrap this and make it so it opens the browser to the releases page instead.

Appreciate your help with this in the meantime! Wish I could be a bit more helpful here, but I can't really do much from outside of the walled garden.

tomitm:
@nerdbeere Can you give this branch one last try? I've updated `electron-packager` and reverted the change in my last comment. Thanks!

tomitm:
Been a while... bunch of `electron-builder` updates in the meantime, some concerning auto-updates.

Will take a another stab at this soon to make sure it works with recent updates. I think this time around, if we can't get macOS updates working, I'll provide a fallback link to the releases until we can work something out.

Would be nice to _finally_ release a 2.0 with auto-updates. 😅

tomitm:
This is now outdated. See #343.

PR #330, opened 2017-05-27
Change the file input to a button (fixes #329)

### LINKED ISSUE #329: Unable to select file in URL Mappings Menu
OS X 10.11.6
James: 1.5.0

When I try to select a file nothing happens. Dragging a file onto the field results in the file being shown (sourcecode)
bildschirmfoto 2017-05-08 um 10 23 31

nerdbeere:
Hey @mitchhentges and @tomitm :wave:,

I made a very basic bugfix for one of the reported issues, maybe someone of you could have a look.

Julian

### DIFF
--- a/src/component/mapping/new-mapping-destination.js
+++ b/src/component/mapping/new-mapping-destination.js
@@ -22,16 +22,9 @@ const NewMappingDestination = (props) => {
 
   let input;
   if (isLocal) {
-    input = {
-      className: 'input-file',
-      disabled: 'disabled',
-      placeholder: 'Choose file',
-      onClick: selectFile
-    };
+    input = <button  className="input-file btn waves-effect waves-light" onClick={selectFile}>Choose file</button>;
   } else {
-    input = {
-      placeholder: 'http(s)://'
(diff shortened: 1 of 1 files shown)


src/component/mapping/new-mapping-destination.js
```diff
@@ -22,24 +22,19 @@ const NewMappingDestination = (props) => {
 
   let input;
   if (isLocal) {
-    input = {
-      className: 'input-file',
-      disabled: 'disabled',
-      placeholder: 'Choose file',
-      onClick: selectFile
-    };
+    input = <button  className="input-file btn waves-effect waves-light" onClick={selectFile}>Choose file</button>;
```
mitchhentges: Can we add `autofocus` to the button, like we do for the `input`? The user can hit `space` or `enter` to run the `onClick` of `<button>`, and it'd be convenient if, regardless of the type of destination, they can just start typing/hitting enter when they hit the page, and things will happen™

Also, you have one too many spaces between `button` and `className`, if you wanna change it while you're at it

PR #340, opened 2017-12-03
Update gulp-sass

mitchhentges:
The super-out-of-date `gulp-sass` that we were using doesn't work on Linux with NodeJS 8.

Soon we're going to need to change from `gulp-sass` (now unmaintained) to `node-sass` with a wrapper (does gulp's `vinyl` cover that?)

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,11 +65,10 @@
     "gulp-json-editor": "^2.2.1",
     "gulp-newer": "^1.2.0",
     "gulp-rename": "^1.2.2",
-    "gulp-sass": "^2.1.1",
+    "gulp-sass": "^3.1.0",
     "gulp-sourcemaps": "^1.6.0",
     "gulp-useref": "^3.0.5",
     "mocha": "^3.0.1",
-    "node-sass": "3.10.1",
     "redux-mock-store": "^1.0.4",
     "referee": "^1.2.0",
     "referee-sinon": "^1.0.3",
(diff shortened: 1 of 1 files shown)


nerdbeere:
We probably could get rid of gulp altogether :)

mitchhentges:
Dang, a lot of hate on the gulp trian 😆 Tomi and I were recently talking about using Gulp vs webpack - there's an awesome ecosystem for building electron apps with Webpack now, so there'd be a large reduction in configuration - but, our current gulp config is still faster than what's been pulled off with webpack so far 😄 

Might be time to modernize so we get those juicy new webpack transform features

nerdbeere:
Don't get me wrong, I hate webpack even more than gulp. 😆 
I'm a fan of simple tooling… still using browserify a lot although it's lacking a lot of features.

Maybe an interesting alternative to webpack: https://parceljs.org/
Another lightweight alternative: https://rollupjs.org/

I think it's worth investing some time in simple and robust tooling.

tomitm:
@nerdbeere Take a look at the [`electron-webpack`](https://github.com/james-proxy/james/tree/electron-webpack) branch. I think you might enjoy how simple it is.

Rollup is [designed for libraries](https://medium.com/webpack/webpack-and-rollup-the-same-but-different-a41ad427058c), so it's not entirely suited for our use case as an application. I saw Parcel come out yesterday, so I'd still wait on that to mature a bit.

PR #343, opened 2017-12-18
Update build / Auto update

tomitm:
It's been ..._a while_... since we've touched `electron-builder` et al. Since then, it seems like a lot has changed. `electron-webpack` removes a lot of build configuration from our project (👋 `gulp`), and `electron-updater` simplifies updates/releases, including direct support for publishing to GitHub releases.

**Highlights**
There's a lot to look over in this branch, sorry, but nothing _too_ drastic.
- Replace `gulp` with `electron-webpack`.
  - This has a bit of convention over configuration, like using the directories common/, main/, and renderer/, which makes sense, realistically.
  - There's also an alias for those directories, so that cleans up `../..` soup a bit.
- Add auto updates again (mostly #284 for the UI bits).
  - Support direct publishing to GitHub releases, so no longer needs a build step to create/update `auto_updater.json` - uses `latest.yml` that's automatically updated to GitHub instead.
- Update to React v16 and React Router v4
  - PropTypes was removed from the `react` package, into to a separate `prop-types` package, which resulted in a bunch of tiny changes
  - RRv4, although controversial initially, is less complicated ditching the route config. Now that it's just components, it's pretty easy to split the bundle by routes for a quicker initial load. Will probably split this further for stuff like the request details view and mapping components.
- Adds `package-lock.json` (npm5)
  - If you ignore that addition, it's actually +637 -596 changes. ¯\\\_(ツ)_/¯
----

**Todo**
- [x] Get tests working again
- [x] Cleanup testing by switching `referee` for `power-assert`, and setup minimal a `webpack.test.config.js` to remove shenanigans
- [x] Cleanup CI by removing xvfb and npm prune/update.
- [x] Figure out why AppVeyor and node-sass aren't getting along.
- [ ] Test auto-updates and verify updates upload to GH releases properly. Currently errors because it gets a 404 on `https://github.com/james-proxy/james/releases/download/v1.5.0/latest.yml` (no longer needs to be in the repo, like the last attempt!)
- [ ] Verify code signing works. Last time we had issues with macOS signing that I'm not sure we entirely resolved?


test.js
```diff
@@ -0,0 +1,51 @@
+const sinon = require('sinon');
+sinon.behavior = {}; // backwards-compat for referee shenanigans
```
tomitm: There's a bunch of shenanigans going on in this file. One of them is the fact `referee` is dated and hasn't kept up with `sinon` changes. Older `sinon` doesn't get along with `webpack` because older `sinon` likes playing around in globals, or something like that. Also, `sinon-as-promised` is no longer needed with the latest `sinon`, but `referee` requires it. 🔥

Other shenanigans are that `electron-webpack` returns a promise for _reasons_, and `mocha` doesn't care for that, so I dug into it's CLI a bit. I went along with the promise shenanigans for build consistency, but I think later I'll write a minimal `webpack.test.config.js` file, to keep things simple again. But I still need this file anyway for now to get `referee` to play nicely.

nerdbeere:
I can help out with the code signing stuff once I got my new mac. Currently I only have access to win10 and ubuntu 17.10.

On Ubuntu, for some reason, I don't see the window when I run `npm start`. It gets rendered in the application switcher. Same thing happens on master so it's not an issue with the changes you introduced.

test.js
nerdbeere: Let's get rid of referee completely and just go with a combination of `mocha`, `sinon` and `power-assert`. I'm very happy with that setup in various other projects I've been working on this year.

tomitm:
I think I have the same issues with [hyper](https://hyper.is/) rendering off screen on startup when I switch from dual monitors to just my laptop. Thought it was specific to hyper, but maybe it's an electron thing in general. There's a config file somewhere that stores the last position. 🤔

tomitm:
Finally got a successful build:
image

Looks like that's a nicely signed package and it automatically uploads to GitHub releases. Wonder what would happen if I created a draft release and changed the version on this branch.

Although, Travis seems to be stalling on the install stage for whatever reason.

nerdbeere:
> I think I have the same issues with hyper rendering off screen on startup when I switch from dual monitors to just my laptop.

are you using gnome? It doesn't happen for me with awesome wm.

nerdbeere:
About travis:

- do we still need this `xvfb` stuff? Maybe we can get rid of it.
- Also we can get rid of `npm prune` and `npm update`. npm should take care of that now.

tomitm:
> are you using gnome? It doesn't happen for me with awesome wm.

Nope, happens on Windows 10.

> do we still need this xvfb stuff? Maybe we can get rid of it.

Ooh you're right! Don't need that anymore now that tests are back to being run under `node` instead of `electron` again. Forgot that's why it was added.

> Also we can get rid of `npm prune` and `npm update`. npm should take care of that now.

Yep, I dropped `npm update` for `npm install` this morning. Need to test locally whether `npm install` with a `package-lock` does pruning properly - I _assume_ it should...

test.js
tomitm: @nerdbeere Shenanigans are out, `power-assert` is in! Do you mind having a look and making sure I changed everything over correctly?

src/renderer/component/footer/footer.js
```diff
@@ -0,0 +1,101 @@
...
+    <UpdateStatus
+      status={props.updateStatus}
+      info={props.updateInfo}
+    />
+  </div>;
+
+Footer.propTypes = {
+  cachingEnabled: PropTypes.bool.isRequired,
+  toggleCaching: PropTypes.func.isRequired,
+  proxyStatus: PropTypes.oneOf([
```
mitchhentges: Should we define these enumerations as their own types?
Otherwise, everywhere where we're referring to `proxyStatus`, we need this whole block of possible values

src/renderer/resources/style/_fonts.scss
```diff
@@ -3,22 +3,22 @@
   font-family: 'Lato';
   font-style: normal;
   font-weight: 300;
-  src: local('Lato Light'), local('Lato-Light'), url(fonts/lato-300.woff2) format('woff2');
+  src: local('Lato Light'), local('Lato-Light'), url('#{$lato-font-path}/lato-300.woff2') format('woff2');
```
mitchhentges: Nice

src/main/auto-update.js
```diff
@@ -0,0 +1,46 @@
...
+  autoUpdater.on('download-progress', (progress) => {
+    updateStatus(constants.UPDATE_DOWNLOADING, progress);
+  });
+  
+  autoUpdater.on('update-downloaded', (info) => {
+    updateStatus(constants.UPDATE_READY, info);
+  });
+  
+  autoUpdater.on('error', (err) => {
+    updateStatus(constants.UPDATE_ERROR, err);
```
mitchhentges: Can we send this error to Sentry? It would be super :( if updates got lost

appveyor.yml
```diff
@@ -8,13 +8,10 @@ cache:
...
 
 install:
-  - ps: Install-Product node 5 x64
+  - ps: Install-Product node 8 x64
   - npm install
-  - npm prune
-  - npm --depth Infinity update
 
 build_script:
-  - npm run test-ci
```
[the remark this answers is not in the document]
tomitm: There's been a `test: false` directive in there for a while, so this seemed like the original intention. There's also a separate `test_script` block we should be using, which `test: false` would've disabled.

Also, I figured now that tests are running under `node` on Linux, it's a bit redundant to run them again on AppVeyor? The reason we're using it is for Windows builds anyway.

src/main/auto-update.js
tomitm: Ooh, will do!

src/renderer/component/footer/footer.js
tomitm: Probably. Any preference where?

src/renderer/resources/style/_fonts.scss
tomitm: Starting to feel the webpack love? 😏

appveyor.yml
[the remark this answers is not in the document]
mitchhentges: I'm (very) mildly worried about tests that fail on Windows and not on Linux.
Besides, if speed-of-build is an issue, the Appveyor and Travis builds run in parallel, so we won't be too negatively affected

src/renderer/component/footer/footer.js
mitchhentges: Probably somewhere in `./src/common`?

appveyor.yml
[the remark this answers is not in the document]
tomitm: Sure, why not. 🤷‍♂️

src/main/auto-update.js
```diff
@@ -0,0 +1,48 @@
...
+  
+  autoUpdater.on('download-progress', (progress) => {
+    updateStatus(constants.UPDATE_DOWNLOADING, progress);
+  });
+  
+  autoUpdater.on('update-downloaded', (info) => {
+    updateStatus(constants.UPDATE_READY, info);
+  });
+  
+  autoUpdater.on('error', (err) => {
```
tomitm: There's an `HttpError` or something that's currently getting thrown here which has a awful `message` that includes a full stacktrace, instead of just the human-readable description it's supposed to be.

PR #345, opened 2017-12-20
Updates travis auth token

mitchhentges:
I regenerated our `GH_TOKEN` for appveyor, but that GH_TOKEN is re-used for both Travis and  Appveyor, so I gotta update both


tomitm:
Let's make sure we're using @james-bot's token for this.

PR #350, opened 2017-12-29
MappingCount component.

### LINKED ISSUE #348: When browser opened from "Home", requests are shown, but titlebar doesn't update
img

Note how "Home" still has the red underline, despite how the panel is showing requests.

*To reproduce:*
1. Start James
2. Click Chrome
* When requests start arriving, the main panel will change to show the requests, but the titlebar will still show "Home" as the active panel

*Acceptance criteria:*
* When a browser is selected:
  * Change panel to show requests being made
  * Update titlebar to show that "Requests" panel is now active

tomitm:
- Split off MappingCount into it's own component - this avoids the `connect` optimization shenanigans that interfere with `react-router`. Fixes #348.
- Change devTools back to open as detached on dev start.
- Switches to hash history (`/#/requests`) to avoid issues with HMR.

### DIFF
--- a/src/main/index.js
+++ b/src/main/index.js
@@ -48,7 +48,7 @@ Raven.context(() => {
     }
     mainWindow.loadURL(rendererURL);
     if (constants.DEV) {
-      mainWindow.webContents.openDevTools();
+      mainWindow.webContents.openDevTools({mode: 'detach'});
     }
 
     mainWindow.webContents.on('did-finish-load', () => {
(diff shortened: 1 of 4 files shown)


src/main/index.js
```diff
@@ -48,7 +48,7 @@ Raven.context(() => {
     }
     mainWindow.loadURL(rendererURL);
     if (constants.DEV) {
-      mainWindow.webContents.openDevTools();
+      mainWindow.webContents.openDevTools({mode: 'detach'});
```
mitchhentges: :clap:

Issue #355, opened 2017-12-30
2.0.0


tomitm:
Todo:
- [x] Figure out how the GitHub auto release upload works
- [x] Test automatic updates between beta.1 to beta.2
- [x] Verify code signing
- [x] Update the changelog

Issues blocking release:
- [x] [Linux/Windows] Restart to update doesn't relaunch - remove behaviour in #360, re-add (without bugs) in #361
- [x] [macOS] Could not get code signature for running application - #357, moved to milestone 2.1

tomitm:
Couple of notes on the GitHub release mechanism:
- It'll create a `draft` release when it detects a new version change
  - The draft release is private to maintainers until we publish it
- If there's already a matching release (say beta.1) and it's published as a `pre-release` and we push a new commit to `master`, since the new commit is considered `draft`, it's a type mismatch and won't replace the upload.
  - If we hadn't of published the draft release, it seems like it would've been updated.
- The draft release gets created with a tag name (e.g., `v2.0.0-beta.1` if package.json used `2.0.0-beta.1`), so we don't need to tag the commit ourselves, publishing does that on the latest commit.
  - Unless you want the tag to not be on the latest commit, then diy - but keep the last sub-bullet in mind.

tomitm:
Couple of issues with update to beta 2:

- Found the change for beta.2 (the version number in the footer) in beta.1 somehow. Build logs don't show that change was uploaded to beta.2 (draft/prerelease mismatch) and GitHub's API timestamps don't show it would've been updated. Weird.
- Draft release gets built on tag, but autoupdate detects the tag as a release, and it fails because latest.yml isn't found. The draft release is still private at this point, which includes latest.yml. Need to change the release to publish immediately, or change how the draft release gets created.
- Update results:
  - Linux: Success. There was an error in Sentry about `app-update.yml` not being found, not sure if it's related, but @mitchhentges reported a success.
  - macOS: Partial failure. Errors with "Could not get code signature for running application" but restarting does successfully update.
  - Windows: Shows an update is available, but `download-progress` event never emits, even though the main process logs "Downloading update from [filename]". No error or anything. Restarts to same version.

mitchhentges:
Everything for `2.0.0` looks pretty good, and there's a [2.1 milestone](https://github.com/james-proxy/james/milestone/2) set up for rounding off the edges of the new auto-update system.

I think it's time to pull the trigger for the release.
[...]

PR #358, opened 2017-12-30
Fix Windows auto-update

tomitm:
- Downgrade `electron-updater` to `2.16.3` to fix the auto-update issue on Windows.
- Update other dependencies.

`electron-updater 2.17.x` seems to have issues with auto-update on Windows. There's been a handful of issues reported about the same symptoms I mentioned in #355 where an update is available, logs that it's downloading, but it never actually downloads or even errors.

I was able to test this by changing my package version to `2.0.0-beta.1`, running `npm run dist`, and installing the newly built package, which then checked GitHub releases for the beta.2 release and successfully downloaded it.

Still have an issue with restart to install, will figure something out in a different PR. It quit without relaunching, but at least we did manage to update to beta.2 without cert errors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "dist:win64": "build --windows --x64"
   },
   "dependencies": {
-    "electron-updater": "^2.17.6",
+    "electron-updater": "2.16.3",
     "font-awesome": "^4.5.0",
     "history": "^4.7.2",
     "hoxy": "^3.2.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -27,7 +27,7 @@
     "dist:win64": "build --windows --x64"
   },
   "dependencies": {
-    "electron-updater": "^2.17.6",
```
mitchhentges: Wait, this is still using the caret, so on NPM 3, `electron-updater@2.17.8` is used.
Should we just un-caret everything? Should we assume nobody is using `npm@3`?

I don't think that there's really that much of a downside to specifying specific versions in our `package.json`

tomitm: Gah, that was unintentional again. 🤦‍♂️

PR #367, opened 2018-01-26
Update dependencies

mitchhentges:


src/renderer/resources/style/main.scss
```diff
@@ -1,5 +1,5 @@
 $fa-font-path: "~font-awesome/fonts/" !default;
-$roboto-font-path: "~materialize-css/fonts/roboto/" !default;
```
mitchhentges: `materialize-css` moved `fonts` to `dist`.
They also have a `css` in `dist`, not sure if there would be advantages to using that, rather than the uncompiled `scss`?

mitchhentges:
I bumped `electron-updater` version so that `electron-builder` would stop whining.
Note that we were holding back `electron-updater`'s version because it had a regression, causing auto-update to fail.
After merging this, we should confirm that it doesn't have any updating issues

Issue #260, opened 2016-08-11
Inspect request and response body


TheBronx:
Is it possible to see the request and response body? For requests that contain text/json this is really useful.

Don't know if this is not implemented or I'm just blind haha. I can only see headers, URL, status code method and time.

nerdbeere:
Hi,
the data is there, unfortunately it's not implemented in the UI.

tomitm:
Some thoughts on the UI for this:

1. Put the body at the bottom of what we currently have
  - Might have a rendering performance cost, but that could be mitigated by hiding the body by default and requiring the user to click a button to show the body.
2. Use tabs, similar to Chrome's devtools
  - Could possibly clash with #76, if we decide to use tabs for displaying multiple requests, but that could be mitigated by using a dropdown to display the multiple requests

Also... I imagine we'd want to show the response raw, rather than attempting to render a preview, right?

nerdbeere:
I would use tabs and basically do it exactly like the devtools are doing things.

avboy72:
Bummer, I finally got a working distro in chrome.... only to find out that I cant see the request/response body.

Raw is fine.  If its JSON would be be a "nice to have" to format it.

ne0c0de:
@avboy72 did you able to see the response body? If yes how? :)

nerdbeere:
It should be a simple fix. All the data you need is here: https://github.com/james-proxy/james/blob/master/src/renderer/component/inspect-request/request-details.js#L39-L56

Feel free to open a PR ❤️

Marty:
I would use tabs too for this wonderful feature :)

carsonip:
Bummer. How could such an important function be missing? At least show it in an ugly way. I would love to see this function.

mitchhentges:
> How could such an important function be missing?

I agree that the feature could be really useful, but sometimes maintainers have other priorities to fulfill.
As always, feel free to submit a PR.

carsonip:
@mitchhentges It is my first time using a proxy for debugging. I tried james first but moved to mitmproxy after realizing james couldn't show response body. I'm not a React ninja but I can mess around and make a PR if I have time.

nerdbeere:
This will be fixed by #377

PR #372, opened 2018-07-09
display response and request body in request details as string

formigarafa:
I bashed the minimum to get the response body out in the request details. People from #260 might want to have look.

### DIFF
--- a/src/common/service/proxy.js
+++ b/src/common/service/proxy.js
@@ -10,8 +10,8 @@ export default class Proxy {
 
     this._proxy = createHoxy();
     this._proxy.intercept('response-sent', this._onResponseSent.bind(this));
-    this._proxy.intercept('request', this._onInterceptRequest.bind(this));
-    this._proxy.intercept('response', this._onInterceptResponse.bind(this));
+    this._proxy.intercept({phase: 'request', as: 'string'}, this._onInterceptRequest.bind(this));
+    this._proxy.intercept({phase: 'response', as: 'string'} , this._onInterceptResponse.bind(this));
   }
 
   _onResponseSent(req) {
@@ -32,6 +32,9 @@ export default class Proxy {
       response.headers.pragma = 'no-cache';
       response.headers['cache-control'] = 'no-cache';
     }
+    response.body = {
+      string: response.string
+    };
   }
 
   _onInterceptRequest(request, response, cycle) {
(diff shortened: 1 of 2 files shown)


src/common/service/proxy.js
```diff
@@ -10,8 +10,8 @@ export default class Proxy {
 
     this._proxy = createHoxy();
     this._proxy.intercept('response-sent', this._onResponseSent.bind(this));
-    this._proxy.intercept('request', this._onInterceptRequest.bind(this));
-    this._proxy.intercept('response', this._onInterceptResponse.bind(this));
+    this._proxy.intercept({phase: 'request', as: 'string'}, this._onInterceptRequest.bind(this));
+    this._proxy.intercept({phase: 'response', as: 'string'} , this._onInterceptResponse.bind(this));
```
mitchhentges: Small nitpick, and the reason why the build failed: there's a space between the `}` and comma here :slightly_smiling_face:

src/renderer/component/inspect-request/request-details.js
```diff
@@ -49,10 +49,18 @@ export default class RequestDetails extends React.Component {
         Request Headers:
         <RequestMetadata metadata={request.request.headers} />
       </section>
+      <section>
+        Request Body:
+        <RequestMetadata metadata={request.request.body} />
```
mitchhentges: Should this be `request.request.body.string`? Isn't `request.request.body` going to be:
```
{
    string: '<body>'
}
```
?

formigarafa: As far I could see metadata need to be an object (Dict). It would probably better to have a specialised component to display body. I don't have the expertise to create one (These were my first lines of code around a react project, and I am aware I did not do any react coding here).

mitchhentges: Oh, shoot, you're right. It looks like `<RequestMetadata>` component is meant to display an object of data, yeah.
Perhaps we can get away with making this section:
```
<section>
  Request Body:
  <RequestMetadata metadata={request.request.body} />
</section>
```

mitchhentges:
Ahaa, looking at this a little further, the test failures seem to be localized to this PR.
Can you run `npm test` and, assuming you have the failing tests locally, resolve them?

nerdbeere:
I'm going to close this one because I opened a new PR with fixed tests (https://github.com/james-proxy/james/pull/377)

PR #374, opened 2018-08-23
Add support for error messages in URL mapping component.

jrmykolyn:


src/renderer/resources/style/components/_errors.scss
```diff
@@ -0,0 +1,21 @@
+.errors {
+    color: firebrick;
+    border: solid 1px firebrick;
+    background: rgba( red, 0.2 );
```
mitchhentges: Nitpick, but would you be able to remove the extra spaces between the braces?
See [_context-menu.scss](https://github.com/james-proxy/james/blob/3dae5adc4bc1afe5cf7f8fa5a79b15317f42799a/src/renderer/resources/style/components/_context-menu.scss#L16) for an example of how it looks elsewhere

test/unit/reducers/url-mappings.js
```diff
@@ -0,0 +1,48 @@
+import assert from 'assert';
```
mitchhentges: Tests! :fire: :clap:

src/renderer/component/errors/errors.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Errors = props =>
+  <div className="errors">
+    <ul className="errors-list">
+      <li className="errors-list__item">
+        {props.errors.map(error => <p>{error}</p>)}
+      </li>
```
tomitm: Could you map the errors to `li` here instead of `p`? This doesn't really make sense semantically otherwise.

src/renderer/resources/style/components/_errors.scss
jrmykolyn: Will do.

src/renderer/component/errors/errors.js
jrmykolyn: Definitely, and good flag.

jrmykolyn:
@mitchhentges @tomitm Thanks for the feedback. I will try and make the updates this evening.

jrmykolyn:
@mitchhentges @tomitm 

Sorry for the delay, work got away from me. I've implemented your feedback. Let me know if you'd like to see any other changes?

src/renderer/component/mapping/new-mapping-destination.js
```diff
@@ -31,6 +33,7 @@ const NewMappingDestination = (props) => {
 
   return <div className="mapping-destination">
     <h1>Enter destination</h1>
+    { errors.length ? <Errors errors={ errors }/> : '' }
```
tomitm: `Errors` should include an empty check so you're not duplicating this logic every time.

jrmykolyn: @tomitm Sounds good.

To confirm: we're talking about moving this logic into the `Errors` component itself?

tomitm: Yep, just a simple `if (!errors.length) return null` within `Errors` should suffice. Then you don't need to worry about it in other components.

jrmykolyn: Sounds good. Will drop that in this evening.

jrmykolyn:
@mitchhentges @tomitm Second round of feedback implemented. Back to you 👍.

Issue #369, opened 2018-03-09
Error: socket hang up


torfeld6:
We're all using macOS in my company and after three to five minutes of using james-proxy with https the following error occurs:

   Error: socket hang up
      at createHangUpError (_http_client.js:302:15)
      at Socket.socketOnEnd (_http_client.js:394:23)
      at emitNone (events.js:91:20)
      at Socket.emit (events.js:188:7)
      at endReadableNT (_stream_readable.js:975:12)
      at _combinedTickCallback (internal/process/next_tick.js:80:11)
      at process._tickDomainCallback (internal/process/next_tick.js:128:9)

Kind Regards
Joshua

torfeld6:
Additional Information: we're not using the browser links, but configured the proxy in the network settings.

![screen shot 2018-03-09 at 16 32 53](https://user-images.githubusercontent.com/661673/37215188-8f7b9310-23b7-11e8-9a0d-e3f8932c35c4.png)

mitchhentges:
Hey, thanks for the bug report - I think that's the same one as [we're seeing internally](https://sentry.fuzzlesoft.ca/share/issue/2d2e152440de4fdd855363a2726fc8e7/).

Two questions:
* Can you remember doing anything in particular to trigger this bug other than just using James for a couple minutes?
* Do you experience any bad behaviour when this bug occurs (I dunno, but maybe something like a request not showing up in James?)

torfeld6:
Hi there,

1. No, I don't. I was opening github.com when it crashed the last time (see log.txt for more information). Is there any other log file to help you?
2. I opened GitHub and there was no response for about 30 seconds before James crashed.

Kind Regards
Joshua

[log.txt](https://github.com/james-proxy/james/files/1797760/log.txt)

mitchhentges:
Oh jeez, it crashed? Sounds pretty serious.
That log file is super helpful, thanks.

I'll take a look when I've got some free time

Defilan:
Hey @torfeld6 and @mitchhentges , I just wanted to hop on and report that this exact same issue is happening on a Fedora 27 system too. 

`  Error: socket hang up
      at createHangUpError (_http_client.js:302:15)
      at Socket.socketOnEnd (_http_client.js:394:23)
      at emitNone (events.js:91:20)
      at Socket.emit (events.js:188:7)
      at endReadableNT (_stream_readable.js:975:12)
      at _combinedTickCallback (internal/process/next_tick.js:80:11)
[...]

PR #376, opened 2018-09-18
Use joxy instead of hoxy

nerdbeere:
This PR fixes https://github.com/james-proxy/james/issues/369 
I did not rename all the code occurrences because I hope we can switch back to hoxy at some point.


nerdbeere:
I actually think that's kind of funny 👍 
Love this dudes attire

PR #377, opened 2018-09-18
Display request/response body (Fixed tests of PR 372)

nerdbeere:
Fixed the failing tests of #372 so we can get this into master.


nerdbeere:
I also refactored the sections a little bit so only sections with actual content are displayed.

With the new `RequestBody` component we could have different visual representations based on the response mime-type. Currently it will be always displayed as text.

PR #381, opened 2018-09-23
Updates to use new Sentry browser/node clients

mitchhentges:
The source mapping issue still exists, but our [`normalizeData` function](https://github.com/james-proxy/james/blob/master/src/common/service/raven.js#L16) isn't as functional as it used to be


mitchhentges:
Nevermind, looks like [we've always been missing sourcemap data](https://sentry.fuzzlesoft.ca/james-proxy/james/issues/257).
I'll assign my :b:oi Tomi as a reviewer, since he set up Sentry the first time around (is there a reason I'm missing for why we should still manipulate the events Sentry sends?)

src/main/index.js
```diff
@@ -1,137 +1,135 @@
...
 import createProxy from './proxy.js';
 import autoUpdater from './auto-update.js';
 
-ravenInit(Raven);
-Raven.context(() => {
-  // Keep a global reference of the window object, if you don't, the window will
-  // be closed automatically when the javascript object is GCed.
-  let mainWindow = null;
+sentryInit(Sentry);
+// Keep a global reference of the window object, if you don't, the window will
```
mitchhentges: Pretty much all of this is just formatting since we didn't need the outside `Raven.context(...)` anymore

PR #387, opened 2018-09-26
Updates electron dependencies

mitchhentges:
Updates dependencies related to building *and packaging* the application.
After merging, we can test updating James (and see what new workarounds `electron-updater` requires these days...)

-----

Also removed the `object-rest-spread` dependency since that's built into babel at this point (or, at least, compilation didn't complain when I removed the now-outdated plugin).

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "babel-eslint": "^10.0.0",
     "babel-loader": "^8.0.4",
     "babel-preset-power-assert": "^3.0.0",
-    "electron": "^2.0.10",
+    "electron": "^3.0.2",
     "electron-builder": "^20.28.4",
     "electron-localshortcut": "^3.0.4",
     "electron-webpack": "^2.3.1",
--- a/package.json
+++ b/package.json
@@ -27,7 +27,6 @@
     "dist:win64": "build --windows --x64"
   },
   "dependencies": {
-    "@babel/preset-env": "^7.1.0",
     "@james-proxy/james-browser-launcher": "^1.3.2",
     "@sentry/browser": "^4.0.2",
     "@sentry/node": "^4.0.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -51,26 +51,29 @@
...
+    "electron": "^3.0.2",
+    "electron-builder": "^20.28.4",
+    "electron-localshortcut": "^3.0.4",
+    "electron-webpack": "^2.3.1",
     "eslint": "^4.16.0",
     "eslint-plugin-react": "^7.6.0",
-    "mocha": "^4.1.0",
-    "mocha-webpack": "^1.0.1",
+    "mocha": "^5.2.0",
+    "mocha-webpack": "^2.0.0-beta.0",
```
mitchhentges: `mocha-webpack@1.x.x` doesn't support `webpack@4`.
I would wait for a full release, but it's lowkey unmaintained.
I'm hoping that the community finds an alternate solution by the time we're worried about this beta usage

mitchhentges:
When this is merged, we can play with this new version of `electron-updater` and see how the new version is working :smile:

package.json
```diff
@@ -51,26 +51,29 @@
...
     "node-sass": "4.7.2",
     "power-assert": "^1.4.4",
     "redux-mock-store": "^1.4.0",
     "rimraf": "^2.6.2",
     "sass-loader": "^6.0.6",
     "sinon": "^4.2.1",
-    "webpack": "^3.10.0",
-    "webpack-sources": "1.0.1"
+    "webpack": "^4.20.2",
+    "webpack-sources": "^1.3.0"
```
tomitm: Did the `webpack-sources` shenanigans get fixed? I remember having to rollback and pin `1.0.1` because of an issue.

tomitm: https://github.com/james-proxy/james/pull/347

Guessing this must've been fixed since there's no build issues. Just FYI in case you forgot. 👍

PR #390, opened 2018-09-30
Uses a key for the array used in request-details

mitchhentges:
The alternative is to wrap the dependent-on-body elements in a `<div>` (or some parent element), but I didn't want to unnecessarily dirty the DOM


mitchhentges:
Just discovered Github's "set reviewer as a *team*" feature. Wow :smile: :clap:

PR #393, opened 2018-10-03
Improves performance by doing deltas over IPC

### LINKED ISSUE #391: Performance: don't send entire request list over IPC
Right now, every time there's a proxy "update", we send the entire list of data over to the renderer process.
Unfortunately, this means that once you hit a couple hundred requests, the time to send the list of requests dwarfs the time between JS ticks, so you get a laggy, unresponsive James.
The solution is to send "deltas" - new requests, and modifying a request with its associated response when it arrives

mitchhentges:
Fixes #391 - major performance benefits after this!

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -87,7 +87,6 @@
     "no-loop-func": 2,               // http://eslint.org/docs/rules/no-loop-func
     "no-multi-str": 2,               // http://eslint.org/docs/rules/no-multi-str
     "no-native-reassign": 2,         // http://eslint.org/docs/rules/no-native-reassign
-    "no-new": 2,                     // http://eslint.org/docs/rules/no-new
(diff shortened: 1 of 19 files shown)


.eslintrc
```diff
@@ -87,7 +87,6 @@
     "no-loop-func": 2,               // http://eslint.org/docs/rules/no-loop-func
     "no-multi-str": 2,               // http://eslint.org/docs/rules/no-multi-str
     "no-native-reassign": 2,         // http://eslint.org/docs/rules/no-native-reassign
-    "no-new": 2,                     // http://eslint.org/docs/rules/no-new
```
mitchhentges: There's a test that needs to create `new Proxy(...)` to test that it wires up callbacks properly.
In that scenario, I don't think it's weird that `new Proxy(...)` has side-effects, so I'm not sure that this lint is benefiting the project :thinking:

PR #402, opened 2018-12-22
Updates build tools

mitchhentges:
Had to due a workaround of https://github.com/webpack/webpack-dev-server/issues/1604, but it's not too bad - I think it's better than pinning to an old version


mitchhentges:
This is supremely out-of-date, so I'll close - when tools _are_ updated, that PR should happen from scratch - this PR is too old to be useful

Issue #405, opened 2018-12-31
App crashes after receiving some packets


jamezrin:
I repeatedly get this exception which crashes the app when running on production mode.

![image](https://user-images.githubusercontent.com/5949554/50560337-24cbae00-0d00-11e9-8152-5f26f31c4362.png)

In the dev console I get
```
hoxy error:  ECONNREFUSED { Error: connect ECONNREFUSED 0.0.0.0:80
      at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1161:14)
    errno: 'ECONNREFUSED',
    code: 'ECONNREFUSED',
    syscall: 'connect',
    address: '0.0.0.0',
    port: 80 }
```

And also 
```
ERROR: Error: connect ECONNREFUSED 0.0.0.0:443
      at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1161:14)
```

I'm running Linux, I tried installing the AppImage build provided and cloning and running it myself.
It might be useful to know that I run a network wide adblocker, PiHole. Which I thought could interfere with James but disabling it doesn't make James to stop crashing, it didn't seem to help.
I have the certificates correctly generated and I'm proxifying the traffic from an Android device to James.

jamezrin:
I worked around this by adding this handler
```
process.on('uncaughtException', (event) => {
  if (event.code === 'ECONNRESET') {
    console.warn('uncaught connection reset'); // eslint-disable-line
  } else if (event.code === 'ECONNREFUSED') {
    console.warn('uncaught refused connection'); // eslint-disable-line
  }
});
```

mitchhentges:
Hmm, investigating this further, the issue doesn't show up in "development mode" because Electron absorbs the error and logs it.

This explains why it wasn't obvious in local development.
In related news, when I tested the `process.on('uncaughtException')` locally, it worked in development mode. However, once a release was created, the issue re-appeared.

I'm not sure why Electron is stopping upon receiving a "fatal error" if we're intercepting it. Maybe we need to do something Electron-specific rather than `process.on(...)`?

Either way, more investigation is needed

kadamgreene:
Is there a new version of James available with this fix in it??

mitchhentges:
Not yet, this issue was trickier than it first seemed.
I haven't had the time to debug further into this, but PRs are always welcome :)

jamezrin:
I think this is fixed in this [PR](https://github.com/greim/hoxy/pull/110).
[...]

PR #407, opened 2019-01-27
Use patched version of hoxy

jamezrin:
The patched version propagates exceptions created by the sockets to the main event emitter, check https://github.com/greim/hoxy/pull/110

The force push is because I accidentally reverted https://github.com/james-proxy/james/pull/406/commits/97993f47d011cf6a89d56dad91c3b1630d1c5735 which would remove the changelog entry for the previous release, so I deleted the handler myself

It works locally, but I think the way I have done this makes no sense (the submodules) so sorry for this, I wanted the commits I created to be visible, but I could've just set my repository on the package.json, but for some reason that wasn't working well when `npm install`ing.

I don't think the PR to hoxy will get merged, it is not maintainted. So maybe the james-proxy can fork it with that fix in it?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "src/proxy"]
+	path = src/proxy
+	url = git@github.com:jamezrin/hoxy.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "src/proxy"]
 	path = src/proxy
-	url = git@github.com:jamezrin/hoxy.git
+	url = https://github.com/jamezrin/hoxy.git
--- a/package-lock.json
+++ b/package-lock.json
@@ -6241,9 +6241,7 @@
       "dev": true
     },
     "hoxy": {
-      "version": "3.3.1",
-      "resolved": "https://registry.npmjs.org/hoxy/-/hoxy-3.3.1.tgz",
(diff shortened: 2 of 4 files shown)


mitchhentges:
@jamezrin thanks for the PR! I'm looking forward to having that [hoxy fix](https://github.com/greim/hoxy/pull/110) be merged. Unfortunately, though that _does_ make `hoxy` behave more correctly, it doesn't seem to be the cause of this crash - if it was, then [this commit](https://github.com/james-proxy/james/commit/97993f47d011cf6a89d56dad91c3b1630d1c5735) should've fixed it.

I checked out your branch locally, but I was still able to reproduce the bug:

1. Check out your branch, `git submodule init && git submodule update` etc etc
2. `npm run clean`
3. `npm run release`
4. I'm on Linux, so I invoke the AppImage created in `binaries` - looks like you're on Linux too, based on [your screenshot](https://github.com/james-proxy/james/issues/405#issue-394974106)
5. [Set up HTTPs proxying](https://github.com/james-proxy/james/wiki#ssl-proxying). This seems to be related (?)
6. Launch Chrome (Our Firefox SSL-proxying config is incorrect I think, so it won't reproduce the bug)
7. This is inconsistent, but I usually continuously open new tabs, go to `google.com`, wait for it to load, and repeat
8. Within ~30s, `james` will crash

If you invoke the AppImage from the terminal, you can see the logs. On my machine in my most-recent test, it crashed with:
```
Error: write EPIPE
    at WriteWrap.afterWrite (net.js:833:14)
```

-----

I don't recall James failing like this in the past, so I'm not sure if this is caused by:
* A change to `hoxy`
* A change to Node or Chrome,
* Some kind of difference in `james` for how we're invoking `hoxy`

Let me know if you can reproduce the problem, and thanks again for digging into this!

